PR #27, opened 2023-08-16
Flake and ghaf-robot wrapper script for running the Robot Framework test suites on Nix/NixOS

mikatammi:
This flake provides a wrapper `ghaf-robot` script to run Robot Framework in an environment where all the required dependency Python modules are present. Additionally few required Python library packages are provided, and also devshell environment is provided.

PR #28, opened 2023-08-19
Support for Kasa Plug KP105

mikatammi:
Still a draft as this needs this another PR as a base


mikatammi:
This is still draft as this needs unmerged #27  as base

PR #36, opened 2023-09-20
Add performance tests

azbeleva:
**!! Can't be merge before sysbench is added to Ghaf.**

Add performance tests for CPU and Memory which doesn't actually have expected results, but execute tests, save results in csv files on Jenkins machine and build plots comparing metrics with previous builds.

Example:
image

### DIFF
--- a/Robot-Framework/config/variables.robot
+++ b/Robot-Framework/config/variables.robot
@@ -4,18 +4,6 @@
 *** Settings ***
 Library    OperatingSystem
 
-*** Variables ***
-${DEVICE}  ${DEVICE}
-
-${SERIAL_PORT}        ${EMPTY}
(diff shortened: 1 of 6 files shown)


Robot-Framework/config/variables.robot
```diff
@@ -15,6 +15,8 @@ ${PASSWORD}           ${EMPTY}
 ${PLUG_USERNAME}      ${EMPTY}
 ${PLUG_PASSWORD}      ${EMPTY}
 ${netvm_service}      ${EMPTY}
```
vjuntunen, line 17: These type of variables should be in Capital letters

azbeleva: Fixed

PR #53, opened 2023-10-31
[SP-3183] Check QSPI version

azbeleva:


mikatammi:
This check won't be enough. I've been having to propose that could you please run the flashing scripts for every testing round?

mikatammi:
I mean, it is good to have this check as well, but it won't be enough to tell if there are changes to the QSPI in the build being tested

mikatammi (approved):

PR #57, opened 2023-11-02
Add LenovoX1 target

azbeleva:
Add tag for LenovoX1 target, refactored Browser test

### DIFF
--- a/Robot-Framework/resources/ssh_keywords.resource
+++ b/Robot-Framework/resources/ssh_keywords.resource
@@ -55,6 +55,38 @@ Connect
     Should Contain    ${output}    ${target_output}
     [Return]          ${connection}
 
+Connect to ghaf host
+    [Documentation]      Open ssh connection to ghaf host
+    Log To Console       Connecting to Ghaf Host
+    ${connection}        Connect
+    Set Global Variable  ${ghaf_host_ssh}    ${connection}
+    [Return]             ${connection}
+
+Connect to netvm
+    [Documentation]      Connect to netvm directly from test run machine, using
+    ...                  jumphost, this allows using standard SSHLibrary
+    ...                  commands, like 'Execute Command'
+    Connect to ghaf host
+    Log To Console       Connecting to NetVM
+    ${connection}        Open Connection   ${NETVM_IP}  port=22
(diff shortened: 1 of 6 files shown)


Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -79,8 +111,9 @@ Verify Systemctl status
 
 Start Chromium
     Log To Console  ${\n}Starting Chromium
-    ${command}=     Set Variable    chromium-browser --enable-features=UseOzonePlatform --ozone-platform=wayland
-    ${command_output}=    Execute Command    nohup ${command} > output.log 2>&1 &
+    Execute Command    sh -c 'cat /etc/xdg/weston/weston.ini | grep chromium | cut -c 6- | sed '\''s/"/\\\"/g'\'' > chromium_launcher'
```
mikatammi, line 114: I wonder what the shell capabilities here are, could we just instead say: `$(cat /etc/xdg/weston/weston.ini | grep chromium | cut -c 6- | sed '\''s/"/\\\"/g'\'')` so we wouldn't need to write intermediate temporary script file.

Also I was thinking are the single and double quotes really matching right in that sed command?

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -79,8 +111,9 @@ Verify Systemctl status
...
+    Execute Command    chmod 755 chromium_launcher
+    Execute Command    nohup sudo -u ${LOGIN} ./chromium_launcher > output.log 2>&1 &
```
mikatammi, line 116: Do we need to use sudo here if we are already logged in as a ghaf user?

mikatammi (commented):
Few things pointed out in comments, otherwise looks good :+1:

Robot-Framework/resources/ssh_keywords.resource
azbeleva: refactored this part of code using your suggestion

azbeleva: also made a method for parsing App path which I plan to use later for other apps

mikatammi (approved):
LGTM, I will still test

Robot-Framework/resources/ssh_keywords.resource
mikatammi: Good, it's always good to refactor things to be more generic :+1:

PR #68, opened 2023-11-22
Jenkinsfiles: Simplify Orin AGX/NX zstd usage

mikatammi:
Instead of writing temporary files first, zstd is now used to directly write the image to disk.

### DIFF
--- a/Jenkinsfiles/hw_test_set
+++ b/Jenkinsfiles/hw_test_set
@@ -94,11 +94,9 @@ pipeline {
                     println("${params.image}")
                     println("${params.device}")
                     if ("${params.device}" == "orin-nx" || "${params.device}" == "orin-agx") {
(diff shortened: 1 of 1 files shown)


vjuntunen (commented):
You tried and verified this to work?

vjuntunen (approved):

mikatammi:
I tried the actual command yes, but I didn't set up another instance of Jenkins to just test this :)

mikatammi:
@vjuntunen So if you can helping on testing the change it would be nice

vjuntunen:
> @vjuntunen So if you can helping on testing the change it would be nice

I have branch where I making some changes for that file about flashing and should I try that change on my branch and add it on that coming PR as well? Otherwise there might be some conflicts after I merge those other changes..

PR #72, opened 2023-11-29
Polarfire SD-MUX flashing

vjuntunen:
Add polarfire flashing via USB-SD-MUX


azbeleva:
Don'r merge it until Polarfire will be set to Production setup

azbeleva (approved):

PR #73, opened 2023-12-04
[SP-3404] Skip browser test if the tested build doesn't contain apps

azbeleva:
RobotFramework will skip browser test if the tested build doesn't contain applications (based on the name of the job):
image

Test run commant must contain the number of the tested build _-v BUILD_ID:$BUILD_ID_ (in Jenkins pipline it must be provided by the parent job)
config directory should contain $BUILD_ID.json
with the name of the job, for example:
```
{
"Job": "nvidia-jetson-orin-agx-debug-nodemoapps-from-x86_64.x86_64-linux"
}
```


vjuntunen (approved):

azbeleva:
Will wait with merging until other changes in main branch will be tested

PR #109, opened 2024-04-21
Add prompt argument to Open Connection calls

mikatammi:
Add prompt argument to SSHLibrary's Open Connection calls. This makes the SSH connection succeed even if it takes more than 0.5 seconds for the actual prompt to appear after SSH connection has been established, and it will also make test faster if it appears earlier.

I wanted to add this because I started to have problems because it started to take more than 0.5 seconds for the actual prompt to appear, and then the connection failed on the "Should Contain" -keyword, because it did not contain the prompt

http://robotframework.org/SSHLibrary/SSHLibrary.html#Login

Here it says "If the [prompt](http://robotframework.org/SSHLibrary/SSHLibrary.html#Prompt) is set, everything until the prompt is read. Otherwise the output is read using the [Read](http://robotframework.org/SSHLibrary/SSHLibrary.html#Read) keyword with the given delay". Delay by default is 0.5


azbeleva (approved):
Thank you! I checked it on all target devices, works fine

PR #115, opened 2024-05-31
Add statistics and PASS/FAIL status based on change in result value to performance tests

leivos-unikie:
Report test PASS/FAIL based on the actual results, not only based on success in running the test commands.

### DIFF
--- a/Robot-Framework/lib/PerformanceDataProcessing.py
+++ b/Robot-Framework/lib/PerformanceDataProcessing.py
@@ -7,6 +7,7 @@ import json
 import matplotlib.pyplot as plt
 import logging
 from robot.api.deco import keyword
+from performance_thresholds import *
 
 
 class PerformanceDataProcessing:
@@ -27,6 +28,8 @@ class PerformanceDataProcessing:
         job = self._get_job_name()
         data_dir = f"../../../Performance_test_results/{job}/"
         os.makedirs(data_dir, exist_ok=True)
+        statistics_dir = f"{data_dir}statistics"
+        os.makedirs(statistics_dir, exist_ok=True)
         return data_dir
 
     def _write_to_csv(self, test_name, data):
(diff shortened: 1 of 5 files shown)


leivos-unikie:
@azbeleva, this is now ready for your review and test runs in the dev test setup

leivos-unikie:
For now the threshold values are hardcoded in Robot-Framework/lib/PerformanceDataProcessing.py. In the future it would be nice to have separate file containing threshold for each performance parameter. I have tuned the threshold values based on production line test result history but I believe we need to tune them more after having some more experience.

leivos-unikie:
This PR does not change the current csv result files and how result data is added to those. Instead it creates a separate statistics file for each test into new "statistics" directory. For now those statistics files are mainly for debugging and following the deviation history. They are overwritten (recalculated based on given threshold values) each time a corresponding test is run.

leivos-unikie:
> For now the threshold values are hardcoded in Robot-Framework/lib/PerformanceDataProcessing.py. In the future it would be nice to have separate file containing threshold for each performance parameter. I have tuned the threshold values based on production line test result history but I believe we need to tune them more after having some more experience.

I moved the threshold settings to Robot-Framework/lib/performance_thresholds.py

azbeleva (approved):

leivos-unikie:
I noticed a bug in assigning correct threshold value for single/multi thread and write/read tests in VMs. Because tests for VMs have different naming convention than the tests for ghaf-host I had to make these changes to PerformanceDataProcessing.py

line 183
```
        # Set threshold for fail depending on test type: single/multi thread
        if "One thread" in test_name or "1thread" in test_name:
            threshold = thresholds['cpu']['single']
        else:
            threshold = thresholds['cpu']['multi']
```

line 296
```
        # Set threshold for fail depending on test type: single/multi thread, write/read
        if "One thread" in test_name or "1thread" in test_name:
            if "rite" in test_name:
                threshold = thresholds['mem']['single']['wr']
            else:
                threshold = thresholds['mem']['single']['rd']
        else:
            if "rite" in test_name:
                threshold = thresholds['mem']['multi']['wr']
            else:
                threshold = thresholds['mem']['multi']['rd']
```

PR #121, opened 2024-07-11
Fix test case fail criteria

leivos-unikie:
Currently any significant deviation (also increase) will cause the test case to fail. With this change test case will not get failed status if the performance improves beyond threshold but only in case the performance decreases.

### DIFF
--- a/Robot-Framework/lib/PerformanceDataProcessing.py
+++ b/Robot-Framework/lib/PerformanceDataProcessing.py
@@ -143,9 +143,11 @@ class PerformanceDataProcessing:
             # Flag indicating significant change in performance value
             flag = 0
 
+            # Check if performance has either decreased or increased significantly
             for i in range(len(d)):
-                if abs(d[i]) > threshold:
-                    # logging.info("Automated check: Performance has changed significantly")
+                if d[i] < -threshold:
+                    flag = -1
+                if d[i] > threshold:
                     flag = 1
 
             stats = self.truncate([mean, pstd] + d + [data_column[-1], data_column[-2], data_column[baseline_start]], 5)
@@ -213,7 +215,7 @@ class PerformanceDataProcessing:
 
                     statistics = self.detect_deviation(data['cpu_events_per_second'], baseline_start, threshold)
 
-                    if statistics['flag'] > 0:
+                    if statistics['flag'] != 0:
                         baseline_start = row_index
 
                     data['statistics'].append(statistics)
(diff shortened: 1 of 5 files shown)


Robot-Framework/lib/PerformanceDataProcessing.py
```diff
@@ -144,8 +144,11 @@ def detect_deviation(self, data_column, baseline_start, threshold):
             flag = 0
 
             for i in range(len(d)):
-                if abs(d[i]) > threshold:
-                    # logging.info("Automated check: Performance has changed significantly")
+                if d[i] < -threshold:
+                    # logging.info("Automated check: Performance has decreased significantly")
```
azbeleva, line 148: Should you uncommentthis logging or remove or make it a comment but not logging?

Robot-Framework/test-suites/log.html
azbeleva, line 1: don't commit log.html. report.html or output.xml

Robot-Framework/test-suites/performance-tests/network.robot
```diff
@@ -32,18 +32,31 @@ TCP speed test
...
+    IF  "${statistics_tx}[flag]" == "1"
+        ${add_msg}     Create improved message  ${statistics_tx}
+        ${pass_msg}=  Set Variable  TX:\n${add_msg}
+    END
+    IF  "${statistics_rx}[flag]" == "1"
+        ${add_msg}     Create improved message  ${statistics_tx}
+        ${pass_msg}=  Set Variable  ${pass_msg}\nRX:\n${add_msg}
+    END
+    IF  "${statistics_tx}[flag]" == "1" or "${statistics_rx}[flag]" == "1"
+        Pass Execution    ${pass_msg}
```
azbeleva, line 57: Why do you use "Pass execution"? This keyword will skip all the rest of the test (teardown also)? Maybe leave a Log message and just pass?

Robot-Framework/lib/PerformanceDataProcessing.py
leivos-unikie: Yep. Removed those and added small comment.

Robot-Framework/test-suites/log.html
leivos-unikie: Did it by accident. Now restored them.

Robot-Framework/test-suites/performance-tests/network.robot
leivos-unikie: I have used "Pass Execution" only in the end of a test so that there should be nothing left in the test to skip. I don't know about teardown though.

The point of using Pass Execution is that I didn't find any other way to pass message to the Message field of a passed test the report table.

leivos-unikie: Custom test status like "IMPROVED" would have been nice but Robot Framework does not allow those.

azbeleva: Yeah it's okay, teardown should work fine

> When used in a test outside setup or teardown, passes that particular test case. Possible test and keyword teardowns are executed.

azbeleva (approved):

PR #133, opened 2024-08-15
Apply changes introduced by ghaf PR724

leivos-unikie:
In ghaf PR724 element-vm is changed to comms-vm and also Slack App is added to the VM. With these fixes the element launch test (kill processes part) survives the changes. Added also launch test for the new Slack App.

In addition created separate lenovo-x1 specific performance threshold parameter for fileio read test because lenovo-x1 fileio rd test results jumped very high few weeks ago causing also variance of the results to increase compared to other targets. So assigning a larger threshold for this specific test/target was appropriate.


leivos-unikie:
This PR should be merged only after https://github.com/tiiuae/ghaf/pull/724 has been merged in.

azbeleva (approved):

leivos-unikie:
ghaf PR#724 went in yesterday evening, so now it is time to merge this.

PR #140, opened 2024-09-19
Automated gui test demo

leivos-unikie:
This PR requires grim and ydotools in ghaf:
https://github.com/tiiuae/ghaf/pull/795
(already merged in)

Proof of concept demonstration how to 
- take screenshots 
- detect image coordinates on screenshot
- move mouse to the desired location
- click mouse
- type text to a text field (log in)

This PR adds new "gui" test suite and new PoC tests `Start and close chromium via GUI on LenovoX1`, `Start Firefox via GUI`, `Boot to login screen` and `Log out and log in`.

`Start Firefox via GUI` demonstrates the principle on other targets than Lenovo-X1. It requires that display is connected to the target device, so this test case should not be added to the ci-test-automation pipeline (no orin/nuc tags added).

Graphical details and icons (.svg) used for image detection are copied from run time ghaf gui-vm. The path variables to these icon libraries are set in the suite Setup. 

Known issues:
- Locate image function is quite prone to errors. For example, running the test multiple times can cause it to fail if the app icon on app menu is left highlighted after the previous launch.
- If some app window is left open behind the app menu it introduces small bias to the background color of the slightly transparent app menu. This causes the locate image to fail if the image includes parts of the background. Cropping only small center part of the launch icon without background can prevent image detection errors. Also introducing tolerance parameter handles this.
- If lenovo-x1 is logged in and left running it will do autolock in 5 min. This will cause gui tests to fail at this point. Lock screen detection could be added in the future. Note: grim works during lock screen but it does not work during Log out screen. 
- Currently only Lenovo-X1 is booting to login screen by default. Orins and NUC boot directly to ghaf desktop. 

UPDATES:
- I added `confidence` parameter which makes image detection more tolerant to small variations, color changes etc.
- Also grayscale parameter would be available for locate function but for now I don't see that necessary in testing ghaf
- Separated gui tests to own suite. Suite setup includes checking if grim is working -> logged in, if not setup does log in using ydotool.
- On lenovo-X1 logs are collected from gui-vm instead of ghaf-host in the Teardown of gui test suite


Robot-Framework/test-suites/bat-tests/gui-tests.robot
maaharko, line 1: Tests shouldn't rely on each other. Now "Close Chromium via GUI on LenovoX1" test is relying on test case "Start Chromium via GUI on LenovoX1". I'd prefer to keep those as one test.

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,102 @@
...
+
+*** Test Cases ***
+
+Start Chromium via GUI on LenovoX1
+    [Documentation]   Start Chromium via GUI test automation and verify related process started
+    [Tags]            bat   SP-T97-1   lenovo-x1
+    Verify service status  range=15  service=microvm@chromium-vm.service  expected_status=active  expected_state=running
+    Start App via GUI on LenovoX1   chromium
+
+Close Chromium via GUI on LenovoX1
```
maaharko, line 28: Tests shouldn't rely on each other. Now "Close Chromium via GUI on LenovoX1" test is relying on test case "Start Chromium via GUI on LenovoX1". I'd prefer to keep those as one test.

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,102 @@
...
+
+*** Settings ***
+Documentation       Testing launching applications
+Force Tags          apps
+Resource            ../../resources/ssh_keywords.resource
+Resource            ../../config/variables.robot
+Resource            ../../resources/common_keywords.resource
+Library             ../../lib/gui_testing.py
+Library             Collections
+Library             BuiltIn
```
maaharko, line 12: It is unnecessary to add BuiltIn library here

Robot-Framework/test-suites/bat-tests/gui-tests.robot
azbeleva: Yeah it can be the one test or you can open or close the app in in different way (by console) , but tests should be independent, that's true

leivos-unikie: App launch and close via GUI has been included into same test.

maaharko (approved):

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,180 @@
...
+    ${mouse_x}  ${mouse_y}  Locate image on screen  ${start_menu}   0.95
+    Execute Command   ydotool mousemove --absolute -x ${mouse_x} -y ${mouse_y}  sudo=True  sudo_password=${PASSWORD}
+    Execute Command   ydotool click 0xC0  sudo=True  sudo_password=${PASSWORD}
+    BuiltIn.Sleep   1
+    ${mouse_x}  ${mouse_y}  Locate image on screen  ${launch_icon}  0.95
+    Execute Command   ydotool mousemove --absolute -x ${mouse_x} -y ${mouse_y}  sudo=True  sudo_password=${PASSWORD}
+    Execute Command   ydotool click 0xC0  sudo=True  sudo_password=${PASSWORD}
+
+    Connect to VM       ${app-vm}
+    Check that the application was started    ${app}
```
azbeleva, line 76: I suggest to add timeout here, by default the range is 2, but let's do it larger

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,180 @@
...
+    ${mouse_x}  ${mouse_y}  Locate image on screen  ../gui-ref-images/start_menu_agx.png  0.95
+    Execute Command   ydotool mousemove --absolute -x ${mouse_x} -y ${mouse_y}  sudo=True  sudo_password=${PASSWORD}
+    Execute Command   ydotool click 0xC0  sudo=True  sudo_password=${PASSWORD}
+    BuiltIn.Sleep   1
+    ${mouse_x}  ${mouse_y}  Locate image on screen  ${launch_icon}  0.95
+    Execute Command   ydotool mousemove --absolute -x ${mouse_x} -y ${mouse_y}  sudo=True  sudo_password=${PASSWORD}
+    Execute Command   ydotool click 0xC0  sudo=True  sudo_password=${PASSWORD}
+
+    BuiltIn.Sleep  1
+    Check that the application was started    ${app}
```
azbeleva, line 124: I guess we don't need additional sleep because it is inside the following keyword, you can increase the number of itterations

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,180 @@
...
+    Connect to VM       ${GUI_VM}
+    Start ydotoold
+
+    ${mouse_x}  ${mouse_y}  Locate image on screen  ${close_button}   0.999
+    Execute Command   ydotool mousemove --absolute -x ${mouse_x} -y ${mouse_y}  sudo=True  sudo_password=${PASSWORD}
+    Execute Command   ydotool click 0xC0  sudo=True  sudo_password=${PASSWORD}
+
+    Connect to VM       ${app-vm}
+    Check that the application is not running    ${app}
```
azbeleva, line 98: this keyword also requires iterations

Robot-Framework/test-suites/bat-tests/gui-tests.robot
leivos-unikie: I put 10

Robot-Framework/test-suites/bat-tests/gui-tests.robot
leivos-unikie: iterations added

Robot-Framework/test-suites/bat-tests/gui-tests.robot
leivos-unikie: Yep

Robot-Framework/resources/common_keywords.resource
```diff
@@ -19,8 +19,13 @@ Check that the application was started
...
     Log To Console       ${app_name} is started
 
+Check that the application is not running
+    [Arguments]          ${app_name}
+    ${found_status}  ${pids}  Run Keyword And Ignore Error   Find pid by name    ${app_name}
+    Should Be Empty   ${pids}
+
 Check If Ping Fails
     [Documentation]  Check that ping is not getting response from host
     # ${out}   Run and Return RC   ping ${DEVICE_IP_ADDRESS} -c 1
```
azbeleva, line 38: commented code

Robot-Framework/resources/common_keywords.resource
```diff
@@ -19,8 +19,22 @@ Check that the application was started
...
+    FOR   ${i}   IN RANGE  ${range}
+        ${keyword_status}  ${pids}  Run Keyword And Ignore Error   Find pid by name    ${app_name}
+        IF  $pids == '${EMPTY}'   BREAK
+        Sleep   1
+    END
+    IF  $pids != '${EMPTY}'
+        Log To Console       ${app_name} is still running
+    END
+    Should Be Empty   ${pids}
+    Log To Console       ${app_name} not running
```
azbeleva, line 34: this part of code doesn't work as supposed to work
image

Suggested:
`    ${pids}=  Set Variable  ${EMPTY}
    FOR   ${i}   IN RANGE  ${range}
        ${keyword_status}  ${pids}  Run Keyword And Ignore Error   Find pid by name    ${app_name}
        ${status}    Run Keyword And Return Status   Should Be Empty  ${pids}
        IF    ${status}    BREAK
        Sleep   1
    END
    Should Be Empty   ${pids}   ${app_name} is still running
    Log To Console    ${app_name} not running`

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,189 @@
...
+    ${mouse_x}  Get From Dictionary   ${coordinates}  x
+    ${mouse_y}  Get From Dictionary   ${coordinates}  y
+    RETURN  ${mouse_x}  ${mouse_y}
+
+Start ydotoold
+    [Documentation]    Start ydotool daemon if it is not already running.
+    ${ydotoold_state}=    Execute Command    sh -c 'ps aux | grep ydotoold | grep -v grep'
+    IF  $ydotoold_state == '${EMPTY}'
+        Log To Console    Starting ydotool daemon
+        Run Keyword And Ignore Error  Execute Command   -b /run/current-system/sw/bin/ydotoold --socket-path /tmp/.ydotool_socket  sudo=True  sudo_password=${PASSWORD}  timeout=3
```
azbeleva, line 178: why do we ignore error? what kind of error can occur which we want to ignore?

Robot-Framework/resources/common_keywords.resource
leivos-unikie: That keyword is from Maarit. I remove the line.

Robot-Framework/resources/common_keywords.resource
leivos-unikie: Yes, that keyword was totally broken. Thanks.

Robot-Framework/test-suites/bat-tests/gui-tests.robot
leivos-unikie: This was a bit tricky part. The thing is that without forcing the timeout (and consequently failing the 'Execute Command' keyword) I could not make the execution to continue. Without timeout robot will just get stuck. When tried manually in gui-vm it requires one hit of enter to return to command prompt. I admit that this is not the best solution but I could not find better for now. Any suggestions?

In the current state this keyword cannot fail even if starting ydotool daemon failed for some reason. If starting the daemon didn't succeed it would show up at next `Execute Command ydotool mousemove...` which would fail. 

It's possible add 
`${ydotoold_state}=    Execute Command    sh -c 'ps aux | grep ydotoold | grep -v grep'` 
`Should Not Be Empty  ${ydotoold_state}  failed to start ydotool daemon`
also to the end of the keyword so that it would actually be able to fail if launching ydotoold didin't succeed.

leivos-unikie: I added the check that launching ydotool daemon succeeded.

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,188 @@
...
+
+    Connect
+    Start ydotoold
+
+    Log To Console    Going to click the app menu icon
+    Locate and click  ../gui-ref-images/start_menu_agx.png  0.95  5
+    Log To Console    Going to click the application launch icon
+    Locate and click  ${launch_icon}  0.95  5
+
+    BuiltIn.Sleep  1
```
azbeleva, line 116: do we still need this sleep?

azbeleva:
don't forget to squash all commits

Robot-Framework/test-suites/bat-tests/gui-tests.robot
leivos-unikie: No we don't

leivos-unikie:
Added ewwbar support also for Orin AGX.

leivos-unikie:
I had to add also automated login because now ghaf boots to login screen.

Robot-Framework/test-suites/bat-tests/gui-tests.robot
```diff
@@ -0,0 +1,228 @@
...
+
+
+*** Variables ***
+@{app_pids}         ${EMPTY}
+@{start_menu}       ../gui-ref-images/ewwbar_start_menu.png
+
+
+*** Test Cases ***
+
+Login on Lenovo-X1
```
maaharko, line 22: This logging to lenovo should be in suite setup, because it is required for next test cases. Also I think the keyword to log in to Lenovo should be somwhere in common keyword resource, since it will be needed in other tests as well

leivos-unikie: There is also the problem that grim is not working during login screen. So we don't have any direct way to verify that login screen is ready to receive input. At boot gui-vm journalctl shows this kind of lines few seconds before the login screen appears:
```
Oct 15 09:52:12 gui-vm seatd[814]: 00:00:05.840 [INFO] [seatd/server.c:145] New client connected (pid: 907, uid: 999, gid: 998)
Oct 15 09:52:12 gui-vm seatd[814]: 00:00:05.840 [INFO] [seatd/seat.c:170] Added client 1 to seat0
Oct 15 09:52:12 gui-vm seatd[814]: 00:00:05.840 [INFO] [seatd/seat.c:480] Opened client 1 on seat0
Oct 15 09:52:14 gui-vm systemd[898]: Created slice User Core Session Slice.
Oct 15 09:52:14 gui-vm systemd[898]: Starting D-Bus User Message Bus...
Oct 15 09:52:14 gui-vm systemd[898]: Started D-Bus User Message Bus.
Oct 15 09:52:14 gui-vm dbus-daemon[920]: [session uid=999 pid=920] Successfully activated service 'org.freedesktop.systemd1'
```

But at the time when login screen appears there are no journalctl logs.

So we are doing login here blindfolded. This requires more attention when adding logout/login test. Anyway, if gui tests are run at later point ghaf should have login screen well ready.

maaharko: `Start and close chromium via GUI on LenovoX1` and  `Start and close Firefox via GUI on Orin AGX `are depending on test `Login on Lenovo-X1`. Normally automated tests should be able to run independently. Therefore this logging should be done already in suite setup

leivos-unikie: I created separate suite for gui tests. Setup of that suite includes log in and it also saves log in/out state after the boot to global variable which is then read in the test case Boot to login screen (to check it didn't boot to desktop directly).

I added log out to the Teardown of gui test suite to render the system into same state (login screen) where it is expected to be when starting the test run.

Robot-Framework/test-suites/gui-tests/__init__.robot
```diff
@@ -0,0 +1,57 @@
...
+    Connect
+    IF  "Lenovo" in "${DEVICE}"
+        Verify service status  range=15  service=microvm@gui-vm.service  expected_status=active  expected_state=running
+        Connect to netvm
+        Connect to VM       ${GUI_VM}
+    END
+    Run journalctl recording
+    Verify logout
+    Log To Console    logged_in_status
+    Log To Console    ${logged_in_status}
```
maaharko, line 31: Is this really working? I mean where is this variable initiated? I can see it getting value in `Verify Logout` but that is run after setup.

In general all suite and global variables should be written in upper letters to differentiate from normal variables

leivos-unikie: I didn't find any problem in using that ${logged_in_status}. It is initiated within the setup (line 29), not after it.

Thanks. I will change it to uppercase.

leivos-unikie: It seems that in this repo there are many lower case global variables. Should I change them all?

Robot-Framework/test-suites/gui-tests/__init__.robot
```diff
@@ -0,0 +1,70 @@
...
+
+    Verify logout
+    Log To Console    LOGGED_IN_STATUS
+    Log To Console    ${LOGGED_IN_STATUS}
+    IF  ${LOGGED_IN_STATUS}
+        Log To Console    Already logged in. Skipping login at suite setup.
+    ELSE
+        Log To Console    Logging in
+        GUI Log in
+    END
```
azbeleva, line 50: make it a one (or several logical) keyword, so setup could be readable

azbeleva: also I would ask you to add tabs - also for readability, something like this:
image

Robot-Framework/test-suites/gui-tests/__init__.robot
```diff
@@ -0,0 +1,70 @@
...
+    Connect
+    IF  "Lenovo" in "${DEVICE}"
+        Connect to netvm
+        Connect to VM       ${GUI_VM}
+    END
+    GUI Log out
+    Log journctl
+    Close All Connections
+
+Run journalctl recording
```
maaharko, line 62: We should avoid defining same keywords in several places. This keyword is defined in bat-suites init-file. Maybe that could be moved to common_keyword.resource

Robot-Framework/test-suites/gui-tests/gui_apps.robot
```diff
@@ -0,0 +1,129 @@
...
+@{app_pids}         ${EMPTY}
+${start_menu}       ./launcher.png
+
+
+*** Test Cases ***
+
+Start and close chromium via GUI on LenovoX1
+    [Documentation]   Start Chromium via GUI test automation and verify related process started
+    ...               Close Chromium via GUI test automation and verify related process stopped
+    [Tags]            bat   SP-T97-1   lenovo-x1
```
azbeleva, line 26: add correct zephyr tag please (without ssrc)

Robot-Framework/test-suites/gui-tests/__init__.robot
```diff
@@ -0,0 +1,64 @@
...
+    IF  "Lenovo" in "${DEVICE}"
+        Verify service status   range=15  service=microvm@gui-vm.service  expected_status=active  expected_state=running
+        Connect to netvm
+        Connect to VM           ${GUI_VM}
+    END
+    Run journalctl recording
+    Save most common icons and paths to icons
+    Verify logout
+    Log To Console              LOGGED_IN_STATUS at start
+    Log To Console              ${LOGGED_IN_STATUS}
```
azbeleva, line 33: suggest
`Log To Console              LOGGED_IN_STATUS at start: ${LOGGED_IN_STATUS}`

Robot-Framework/resources/gui_keywords.resource
```diff
@@ -0,0 +1,137 @@
...
+${start_menu}       ./launcher.png
+
+*** Keywords ***
+
+GUI Log in
+    [Documentation]   Login and verify that task bar is available
+    Start ydotoold
+    Log To Console    Typing username and password to login
+    Type string and press enter  ghaf
+    Type string and press enter  ghaf
```
azbeleva, line 20: use global variables ${LOGIN} and ${PASSWORD}

Robot-Framework/resources/gui_keywords.resource
```diff
@@ -0,0 +1,137 @@
...
+        Locate and click  ./icon.png  0.95  5
+        Get icon  ghaf-artwork  logout  crop=0  background=black
+        Locate and click  ./icon.png  0.95  5
+    END
+    Verify logout
+    Run Keyword If    not ${LOGGED_IN_STATUS}  FAIL   Log out failed. Desktop is still running.
+
+Type string and press enter
+    [Arguments]   ${string}
+    Log To Console    Typing ${string}
```
azbeleva, line 54: let's not log it here at least for a while

Robot-Framework/resources/gui_keywords.resource
```diff
@@ -0,0 +1,137 @@
...
+    [Arguments]        ${log_out_icon}=./logout.png
+    Start ydotoold
+    Log To Console    Going to click the app menu icon
+    Locate and click  ${start_menu}  0.95  5
+    Log To Console    Checking if logout icon is in app menu
+    ${login_icon_found_status}  ${msg}  Run Keyword And Ignore Error  Get icon  app  system-log-out  crop=20  filename=logout
+    ${locate_status}=  Set Variable  FAIL
+    IF  $login_icon_found_status == "PASS"
+        ${locate_status}  ${msg}  Run Keyword And Ignore Error  Locate image on screen  ./logout.png
+        # Log To Console  ${locate_status}
```
azbeleva, line 35: comment

Robot-Framework/test-suites/gui-tests/gui_apps.robot
```diff
@@ -0,0 +1,129 @@
...
+
+    [Teardown]    Run Keywords    Connect to VM     ${GUI_VM}
+    ...           AND             Move cursor to corner
+
+Close app via GUI on LenovoX1
+    [Documentation]    Close Application via GUI test automation and verify related process stopped
+    [Arguments]        ${app-vm}
+    ...                ${app}
+    ...                ${close_button}=./window-close.png
+    # ...                ${close_button}=../gui-ref-images/${app}/close_button.png
```
azbeleva, line 72: comment

flake.nix
```diff
@@ -42,6 +42,8 @@
       devShell = pkgs.mkShell {
         buildInputs = with pkgs; [
           iperf
+          file
+          # How to add imagemagick
```
avnik, line 46: `imagemagick` here is enough for get im in devshell

Robot-Framework/test-suites/gui-tests/__init__.robot
```diff
@@ -0,0 +1,64 @@
...
+    END
+    GUI Log out
+    Log journctl
+    Close All Connections
+
+Save most common icons and paths to icons
+    [Documentation]         Save those icons by name which will be used in multiple test cases
+    ...                     Śave paths to icon packs in gui-vm nix store
+    Log To Console          Saving path to icon-pack
+    ${app_icon_pack_path}   Execute Command   echo /nix/store/$(ls /nix/store | grep icon-pack | tail -1)
```
avnik, line 55: Running `$(ls /nix/store ...)` is pretty slow on big stores, also is indeterminism.  Could we provide path via devshell's environment variable?

Robot-Framework/test-suites/gui-tests/__init__.robot
```diff
@@ -0,0 +1,64 @@
...
+    Set Global Variable     ${APP_ICON_PATH}  ${app_icon_pack_path}
+    Log To Console          ${APP_ICON_PATH}
+    Log To Console          Saving path to ghaf-artwork icons
+    ${ghaf_artwork_path}    Execute Command   echo /nix/store/$(ls /nix/store | grep ghaf-artwork | tail -1)/icons
+    Set Global Variable     ${ARTWORK_PATH}  ${ghaf_artwork_path}
+    Log To Console          ${ARTWORK_PATH}
+    Log To Console          Saving gui icons
+    Get icon                ghaf-artwork  launcher  crop=0  background=black  filename=launcher
+    Get icon                /run/current-system/sw/share/icons/Adwaita/symbolic/ui  window-close-symbolic  crop=0  filename=window-close
```
avnik, line 64: One more, could we pass adwaita as environment variable?

Robot-Framework/lib/gui_testing.py
```diff
@@ -0,0 +1,39 @@
...
+    scaling_factor = 2
+    mouse_coordinates = {
+        'x': int(coordinates[0] / scaling_factor),
+        'y': int(coordinates[1] / scaling_factor)
+    }
+    return mouse_coordinates
+
+def convert_app_icon(crop, background, filename):
+    if background != "none":
+        subprocess.run(['magick', '-background', background, 'icon.svg', '-gravity', 'center', '-extent',
```
avnik, line 34: you perhaps probably want `convert` executable, because imagemagick haven't single `magick` executable, but bunch of tools like `convert`, `conjure`, whatever else

avnik (changes requested):
I suggest some changes related imagemagick, and add more notes

flake.nix
leivos-unikie: Indeed very simple. I had already tried that but it didn't work then. I don't know what I did differently. Now it's working.

Robot-Framework/lib/gui_testing.py
leivos-unikie: In IMv7 convert is deprecated, the recommended use for convert is just
`magick inputfile outputfile`

Robot-Framework/test-suites/gui-tests/__init__.robot
leivos-unikie: Currently the execution time of both of those `ls` commands is well below 0.04s. I don't see any problem in that. How fast should it be? 

Do you mean using some already existing environment variable or creating one?

I thought that the nix store path to that icon-pack and ghaf-artwork might be changing between builds, so we cannot hardcode a variable for them.

Robot-Framework/test-suites/gui-tests/__init__.robot
leivos-unikie: I can make a variable for that adwaita path. 

Does it matter whether I am using dev shell environment variable or robot framework variable?

Robot-Framework/test-suites/gui-tests/gui_apps.robot
leivos-unikie: removed

Robot-Framework/resources/gui_keywords.resource
leivos-unikie: debugging, reemoved

Robot-Framework/resources/gui_keywords.resource
leivos-unikie: yes. thanks.

Robot-Framework/test-suites/gui-tests/gui_apps.robot
leivos-unikie: changed

Robot-Framework/test-suites/gui-tests/__init__.robot
leivos-unikie: it's more beautiful now

Robot-Framework/test-suites/gui-tests/gui_login.robot
```diff
@@ -0,0 +1,28 @@
...
+Resource            ../../config/variables.robot
+Resource            ../../resources/gui_keywords.resource
+Suite Teardown      Close All Connections
+
+
+*** Test Cases ***
+
+Boot to login screen
+    [Documentation]   Check that device booted to login screen
+    [Tags]            bat   lenovo-x1   SP-T2
```
azbeleva, line 17: remove 'bat' tag please

Robot-Framework/test-suites/gui-tests/gui_login.robot
```diff
@@ -0,0 +1,28 @@
...
+    Run Keyword If    ${LOGGED_IN_STATUS}  FAIL  Desktop was detected at setup. Device failed to boot to login screen.
+
+Log out and log in
+    [Tags]            bat   lenovo-x1   SP-T149
```
azbeleva, line 21: remove 'bat' tag please

Robot-Framework/test-suites/gui-tests/gui_login.robot
azbeleva, line 1: I have doubt about this file. Basically you are doing log in in setup and log out in teardown, if one of the steps fails - test fails. So if there are some problems with it - you will see it. Then you have SP-T149 test case where you are doing the same actions one more time in reverse order. And I also think that SP-T2 is kind of strange, because the tested feature is included in setup. I can't say now what is the better way, but I am not sure about current implementation

leivos-unikie: Maarit suggested that login should be added to the suite setup because all gui tests depend on login. I agreed to this but the consequence is that I didn't find any other way to check 'boot to log screen' than to include it to suite setup before the initial login. However, this test can be removed if it is not considered necessary.

The drawback of failing logout in Teardown is that it infects all gui test cases with fail status. But if login is in Setup I thought we should include logout to Teardown to leave the system to the same state where it is supposed to be when starting gui test suite. 

But if we remove 'Boot to login screen' then maybe we can also remove logout from Teardown and change SP-T149 to test only logout. But then it should be ensured that this test is run last in the gui test suite.

What do you think?

leivos-unikie: Actually it now came to my mind that if we add a test case for pressing the "Reboot" button in the power menu that check for booting into login screen could be moved as part of "gui reboot" test case and the login screen state does not need to be saved in the suite Setup.

azbeleva: I like your idea with reboot, but it will check two things (reboot button and login screen), but it's okay for now

leivos-unikie: Ok. I will add that. Booting into login screen can be considered as part of successful reboot so I think it is justified.

leivos-unikie: Although login is already in suite setup (and whole suite will fail if login fails) I added Login also as a separate test case between 'GUI Reboot' and Logout since there has to be login anyway.

leivos-unikie:
When the number of GUI tests increase new `Connect to VM if not already connected` keyword can save testing time significantly by reducing the unnecessary re-connection cycles.

That keyword could potentially speed up also some currently existing test cases.

Robot-Framework/test-suites/gui-tests/gui_power_options.robot
```diff
@@ -0,0 +1,63 @@
...
+Suite Teardown      Close All Connections
+
+
+*** Test Cases ***
+
+GUI Reboot
+    [Documentation]   Reboot the device via GUI reboot icon.
+    ...               Check that it shuts down. Check that it turns on and boots to login screen.
+    [Tags]            lenovo-x1   SP-T208-1
+    Connect to VM if not already connected  gui-vm
```
azbeleva, line 20: I don't think you should actually make such checks, it's the beginning of the test and you should close connections in the end of each test and create new in new test case.
If you need to switch between connections inside of one test you can save the connection as a variable and use it in keyword 'Switch Connection'

leivos-unikie: Why not to use same connection over multiple app launch test cases?

I looked the app launch tests in app.robot as a reference. They are not closing connections in the end of each test case (only the bat test suite has Close All Connections in the Common Teardown) but this causes open ssh connections to accumulate (on ghaf-host and net-vm). This can be monitored by running `watch who` in ghaf-host during test run. I just thought, instead of opening new ssh connection for each test case, to use the existing one, to prevent accumulating connections and to save some time.

Do you see some risk in using the same ssh connection over multiple tests?

azbeleva: let's reconsider that later, for now merge how it is

azbeleva (approved):

PR #151, opened 2024-10-15
Refactor connections keywords

azbeleva:


Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -367,7 +367,8 @@ Get wifi IP
 
 Check if ssh is ready on device
     [Arguments]    ${timeout}=30
-    ${is_ready} =  Set Variable    False
+    ${is_ready}    Set Variable    False
```
maaharko, line 370: Shouldn't this be quite simpler if we use  Wait Until Keyword Succeeds here? E.g
```
Check if ssh is ready on device

  ${rc}  	${output}	 Run And Return Rc And Output   nc -zvw3 ${DEVICE_IP_ADDRESS} 22
  ${status}    Run Keyword And Return Status
  ... Should Be Equal As Integers ${rc} 0
```

....

`Wait Until Keyword Succeeds  ${timeout}  1s  Check if ssh is ready on device`

azbeleva: This keyword does not fail, but return the status, it was done on purpose. we can make failable but it will require refactoring in other cases with using keywords which ignore errors

maaharko: ok

maaharko (approved):

PR #153, opened 2024-10-16
Update Boot time test

maaharko:
Update test to search different word from journalctl


leivos-unikie (approved):
I would suggest changing
`Time since reboot to desktop avaialble`
to
`Time since reboot to login screen`

azbeleva (approved):

PR #160, opened 2024-10-30
Modify memory status test

maaharko:
Modify memory test to accept floats and integers.
Note, commit name is incorrect, this is about memory status check


azbeleva:
it’s possible to change the commit message, and PR name too, let’s do that for clean commit history

azbeleva (approved):

PR #166, opened 2024-11-07
Performance tests

maaharko:
Add more performance tests

### DIFF
--- a/Robot-Framework/lib/PerformanceDataProcessing.py
+++ b/Robot-Framework/lib/PerformanceDataProcessing.py
@@ -83,10 +83,10 @@ class PerformanceDataProcessing:
 
     @keyword
     def write_speed_to_csv(self, test_name, speed_data):
-        data = [self.build_number,
-                speed_data['tx'],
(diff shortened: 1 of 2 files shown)


leivos-unikie:
I think the old iperf test cases (TCP/UDP speed test) could be removed as they seem to be doing the same thing as the new test cases "Measure TCP/UDP Throughput Small Packets", just lacking the reversed direction.

leivos-unikie:
Currently Robot-Framework/lib/performance_thresholds.py has only one threshold parameter (10) for iperf tests which is now applied to all of the new iperf test cases. This can be totally fine. At least those old network test results have been extremely stable. If it is noticed later on that some of these new tests have higher deviation in results then new thresholds should be added.

maaharko:
> I think the old iperf test cases (TCP/UDP speed test) could be removed as they seem to be doing the same thing as the new test cases "Measure TCP/UDP Throughput Small Packets", just lacking the reversed direction.

I agree. But I would like to run new tests for a while before removing old ones

PR #181, opened 2024-12-03
Add video tests

maaharko:
Video tests
- Checking that camera is available only in business-vm
- Checking that video clip can be recorded, and some basic checks for video file is done

### DIFF
--- a/Robot-Framework/test-suites/video-tests/video.robot
+++ b/Robot-Framework/test-suites/video-tests/video.robot
@@ -1,0 +1,83 @@
+# SPDX-FileCopyrightText: 2022-2024 Technology Innovation Institute (TII)
+# SPDX-License-Identifier: Apache-2.0
+
+*** Settings ***
+Documentation       Testing camera applicaton
+Force Tags          video  lenovo-x1
+Resource            ../../__framework__.resource
+Resource            ../../resources/serial_keywords.resource
+Resource            ../../resources/common_keywords.resource
+Resource            ../../resources/connection_keywords.resource
+Resource            ../../resources/ssh_keywords.resource
+Resource            ../../config/variables.robot
+Suite Setup         Initialize Variables And Connect
+Suite Teardown      Close All Connections
+Test Timeout        2 minutes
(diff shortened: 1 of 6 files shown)


Robot-Framework/test-suites/video-tests/video.robot
```diff
@@ -0,0 +1,83 @@
...
+Resource            ../../config/variables.robot
+Suite Setup         Initialize Variables And Connect
+Suite Teardown      Close All Connections
+Test Timeout        2 minutes
+
+
+*** Test Cases ***
+Check Camera Application
+    [Documentation]  Check that camera application is available in business-vm and not in other vm
+    [Tags]  SSRCSP-T235
```
leivos-unikie, line 21: I think it was agreed to have the shorter form of ids in tags -> SP-T235

Robot-Framework/test-suites/video-tests/video.robot
```diff
@@ -0,0 +1,83 @@
...
+    FOR  ${vm}  IN  @{VMS}
+        Connect to VM       ${vm}
+        ${out}  Execute Command  v4l2-ctl --list-devices
+        Log  ${out}
+        IF  '${vm}' == '${BUSINESS_VM}'  Should Contain  ${out}  /dev/video  ELSE  Should Not Contain  ${out}  /dev/video
+    END
+
+Record Video With Camera
+    [Documentation]  Start Camera application and record short video
+    [Tags]  SSRCSP-T236
```
leivos-unikie, line 32: -> SP-T236

leivos-unikie:
If I run Record Video With Camera test with camera lid on the test fails with message
'2073600 < (2073601-1)' should be true.
in the message field which means that all pixels of the captured image have same color. Could the fail message be more clear? For example: "Capturing video succeeded but real content was not detected. (Check the camera lid.)"

Also if the video file is empty the test fails with message
'0 > 0' should be true.
Could this also be more informative?

azbeleva (approved):

PR #197, opened 2024-12-18
Fix video tests

maaharko:
Create video files to /tmp folder


milva-unikie (approved):
Tested that this works (at least locally)

PR #200, opened 2024-12-30
Test IP spoofing

leivos-unikie:


milva-unikie:
I tested that this works as intended on X1. Someone else should probably review the code before merging.

Robot-Framework/test-suites/security-tests/__init__.robot
```diff
@@ -0,0 +1,36 @@
...
+Suite Setup         Common Setup
+Suite Teardown      Common Teardown
+
+*** Variables ***
+
+${connection}       ${NONE}
+
+*** Keywords ***
+
+Common Setup
```
azbeleva, line 17: let's choose some more reasonable name, not "common", we changed it in other init files too

Robot-Framework/test-suites/security-tests/security_tests.robot
```diff
@@ -0,0 +1,81 @@
...
+    [Documentation]   Test if it's possible to steal packets via ip spoofing
+    [Tags]            SP-T128   lenovo-x1
+
+    # Prepare netcat server script
+    Connect to netvm
+    Check if ssh is ready on vm    ${GALA_VM}
+    Connect to VM                  ${GALA_VM}
+    ${ip_gala}                     Get Virtual Network Interface IP
+    Put File                       security-tests/nc_server   /tmp
+    Execute Command                chmod 777 /tmp/nc_server
```
azbeleva, line 28: can we put those to keywords and make test case shorter and easier to read?
for example create keywords for blocks of code which are already separated using comments:
Prepare netcat server script
Prepare netcat client script
Prepare netcat stealer script
Launch the test scripts
Check the result files

Robot-Framework/test-suites/security-tests/__init__.robot
leivos-unikie: I simplified the init, removed separate setup keywords and used existing.

Logging ghaf-host journalctl is for no use in this test.

Robot-Framework/test-suites/security-tests/security_tests.robot
leivos-unikie: Refactored and generalized so that VMs can be easily changed if necessary.

azbeleva (approved):

PR #204, opened 2025-01-07
Fix BAT teardown for not Lenovo

azbeleva:


leivos-unikie (approved):

milva-unikie (approved):
Tested with AGX. The time synchronization test case failed for some reason, but other than that everything seems to be good.

PR #207, opened 2025-01-09
Integrate video, audio and logging tests to bat suite

leivos-unikie:


leivos-unikie:
Tested running bat suite with https://github.com/tiiuae/ghaf/pull/917 ghaf image. All tests passed except 
Check Grafana logs
Wifi passthrought into NetVM
which require WiFi (and I didn't configure wifi)

Audio tests pass if they are run after logging in only once after reboot. If there has been log out and log in after last boot audio tests fail. It is not yet clear why this happens. Basically bat suite is run right after boot test in main pipeline but if there are tests runs without reboot between audio tests would fail.

azbeleva (approved):

PR #213, opened 2025-01-22
Add relay_boot_test.robot

vjuntunen:
Add relay_boot_test.robot and needed keywords
to use relay for boot device by cutting down power.

Add KMTronic relay board control and status python scripts and README. Add also Robot framework library for KMTronic control.

### DIFF
--- a/Robot-Framework/config/variables.robot
+++ b/Robot-Framework/config/variables.robot
@@ -23,6 +23,8 @@ Set Variables
 
     ${config}=     Read Config
     Set Global Variable  ${SERIAL_PORT}        ${config['addresses']['${DEVICE}']['serial_port']}
+    Set Global Variable  ${RELAY_SERIAL_PORT}  ${config['addresses']['relay_serial_port']}
+    Set Global Variable  ${RELAY_NUMBER}       ${config['addresses']['${DEVICE}']['relay_number']}
     Set Global Variable  ${DEVICE_IP_ADDRESS}  ${config['addresses']['${DEVICE}']['device_ip_address']}
     Set Global Variable  ${SOCKET_IP_ADDRESS}  ${config['addresses']['${DEVICE}']['socket_ip_address']}
     Set Global Variable  ${PLUG_TYPE}          ${config['addresses']['${DEVICE}']['plug_type']}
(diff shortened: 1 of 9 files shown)


Robot-Framework/config/variables.robot
```diff
@@ -23,6 +23,8 @@ Set Variables
 
     ${config}=     Read Config
     Set Global Variable  ${SERIAL_PORT}        ${config['addresses']['${DEVICE}']['serial_port']}
+    Set Global Variable  ${RELAY_SERIAL_PORT}  ${config['addresses']['relay_serial_port']}
+    Set Global Variable  ${RELAY_NUMBER}       ${config['addresses']['${DEVICE}']['relay_number']}
```
azbeleva, line 27: won't it fail for setup where there are no relay boards yet?

Robot-Framework/test-suites/boot-test/relay_boot_test.robot
```diff
@@ -0,0 +1,123 @@
...
+    END
+    IF  "${CONNECTION_TYPE}" == "ssh"
+        Connect
+        Verify service status   service=init.scope
+    ELSE IF  "${CONNECTION_TYPE}" == "serial"
+        Verify init.scope status via serial
+    END
+    [Teardown]   Test Teardown
+
+Verify booting LenovoX1
```
azbeleva, line 41: is this test case just a copy paste? should it be in this file if Lenovo is not controlled by relay?

Robot-Framework/config/variables.robot
vjuntunen: Oh you're right about that.. Maybe need to add those properties with empty value for now.

Robot-Framework/test-suites/boot-test/relay_boot_test.robot
vjuntunen: Yes it does if we use relay-boot tag and also want to have that x1 in same set. We only give one boot tag in our pipelines. And when we have relay in use in every setup that will be just 'boot'. This is just temporary to store it to repo and use it where needed.

vjuntunen: It is simple and clear solution for now. Then we can just remove old boot test suite and replace it with this.

Robot-Framework/config/variables.robot
leivos-unikie: I added IP of measurement agent previously like this to variables.robot

`Run Keyword And Ignore Error  Set Global Variable  ${RPI_IP_ADDRESS}  ${config['addresses']['measurement_agent']['device_ip_address']}
`

vjuntunen: I will deploy that change before merge this one.

azbeleva (approved):

PR #215, opened 2025-01-24
Add scaling reset to GUI Log in

milva-unikie:
1. Add a reset for automatic scaling (https://github.com/tiiuae/ghaf/pull/960). Without reset tests are not able to confirm logged in state and almost all of them fail. Works also with current mainline version so can be merged in before #960.
2. Fix Stop swayidle. I noticed that it was not working anymore and it has probably been broken since user accounts were added.

### DIFF
--- a/Robot-Framework/resources/gui_keywords.resource
+++ b/Robot-Framework/resources/gui_keywords.resource
@@ -28,9 +28,9 @@ Log in via GUI
         Type string and press enter  ${USER_LOGIN}
         Type string and press enter  ${USER_PASSWORD}
     END
+    Try to reset scaling
     Verify login
-    Log To Console    Disabling automated lock and suspend
-    Execute Command   systemctl --user stop swayidle
+    Stop swayidle
 
 Log out
     [Documentation]   Log out and optionally verify that desktop is not available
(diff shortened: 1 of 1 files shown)


leivos-unikie:
> 1. Add a reset for automatic scaling ([Desktop: Automatic scaling on HiDPI displays ghaf#960](https://github.com/tiiuae/ghaf/pull/960)). Without reset tests are not able to confirm logged in state and almost all of them fail. Works also with current mainline version so can be merged in before #960.
> 
>     2. Fix Stop swayidle. I noticed that it was not working anymore and it has probably been broken since user accounts were added.

1. Have you run bat and gui test suites with ghaf mainline and #960? And they passed?

2. I thought that also previously the ssh connection would have been connected to gui-vm as testuser before running that `systemctl --user stop swayidle` command. `Verify login` calls `Locate image on screen` which logs in to gui-vm as testuser. But anyway it's nice to have separate keyword for that to ensure.

milva-unikie:
> 1. Have you run bat and gui test suites with ghaf mainline and #960? And they passed?

Yes, yes and yes. 

> 2. I thought that also previously the ssh connection would have been connected to gui-vm as testuser before running that `systemctl --user stop swayidle` command. `Verify login` calls `Locate image on screen` which logs in to gui-vm as testuser. But anyway it's nice to have separate keyword for that to ensure.

It was definitely not working, I checked it multiple times and swayidle was not getting stopped. Verify login does not return the connection, so I believe the command was run with ghaf-account.

leivos-unikie (approved):

PR #219, opened 2025-01-28
Allow setting more variables via command line

leivos-unikie:
The purpose of this PR is to bring more flexibility into manual use of ci-test-automation and to make local smoke testing faster. Fully automated setups should not be affected.

CONFIG_PATH (path to test_config.json) can be given via command line. If not given it defaults to the current path. 

If CONFIG_PATH is set to `None` reading config file variables will be ignored which allows setting target IP address from command line. This makes local testing more straightforward, not having to create or modify test_config.json.

Prevents Set Variables from overwriting password (given via cmd line) if /run/secrets/dut-pass does not exist.

Custom test user name and password are now allowed. They can be stored to /etc/secrets/testuser and /etc/secrets/testpw. If those files don't exist test user credentials default to testuser/testpw. 

**Steps to run smoke tests locally from ghaf-host:**
```
Open terminal and ssh to ghaf-host
cd /home
sudo su
git clone https://github.com/leivos-unikie/ci-test-automation.git
cd ci-test-automation/Robot-Framework/test-suites
nix develop
git checkout optional_config
robot -v CONFIG_PATH:None -v DISABLE_LOGOUT:true -v DEVICE:Lenovo-X1 -v DEVICE_IP_ADDRESS:192.168.100.1 -v PASSWORD:ghaf -i lenovo-x1ANDpre-merge ./
```

I ran also with
`robot -v CONFIG_PATH:None -v DEVICE:Lenovo-X1 -v DEVICE_IP_ADDRESS:192.168.100.1 -v PASSWORD:ghaf -i lenovo-x1ANDbat -i lenovo-x1ANDgui-apps ./`
after setting wifi credentials to /run/secrets and all test cases passed except timesync test (in current state it assumes unchanged time on the machine running robot framework).

If running the tests from separate machine just set the target IP to DEVICE_IP_ADDRESS.

If running tests locally from ghaf-host of Lenovo-X1 the result files can be sent to chrome-vm after the tests have finished by `python ../lib/send_report.py password` and then opened with chrome app by typing /tmp/report.html to the url field.

### DIFF
--- a/Robot-Framework/config/variables.robot
+++ b/Robot-Framework/config/variables.robot
@@ -14,6 +14,7 @@ ${TEST_WIFI_SSID}  ${EMPTY}
 ${TEST_WIFI_PSWD}  ${EMPTY}
 ${DEVICE_TYPE}     ${EMPTY}
 ${LOGIN}           ghaf
+${CONFIG_PATH}     ../config/test_config.json
 
 
 *** Keywords ***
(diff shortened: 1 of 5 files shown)


milva-unikie:
Great start! I tested this by running the tests from ghaf host on Lenovo-X1. Few comments:
- ssh command should be `ssh ghaf@ghaf-host-debug` and `nix develop` is missing
- If you have created your user as something else than testuser tests won't work unless you go and change the username/password in variables.robot
- At the end tests log out and results in terminal are lost
- No easy way to check report.html

leivos-unikie:
Thanks for trying this out. My suggestions to these.

>     * ssh command should be `ssh ghaf@ghaf-host-debug` and `nix develop` is missing
The first step was meant generally somehow connecting to ghaf-host either from gui-vm or from separate machine over network.

>     * If you have created your user as something else than testuser tests won't work unless you go and change the username/password in variables.robot
Or is it replacing your previously set user with the testuser?

>     * At the end tests log out and results in terminal are lost
We could skip all automated Logouts if CONFIG_PATH is set to None.  

>     * No easy way to check report.html
I checked that is actually possible to open the report with chrome if first copying the report files to chrome-vm:/tmp and giving them appuser:users ownership. Or then copying to other machine for study... but this is not very smooth. I try to automate sending the results for chrome to read.

milva-unikie (approved):
Thanks for the improvements! Now this works great. I tested running both pre-merge-tests and bat-tests in Ghaf from ghaf-host and then checked the results in Chrome.

> Or is it replacing your previously set user with the testuser?

It won't replace the user. If the username/password are different the tests just won't work.

(As a note about the documentation: `python ../lib/send_report.py` also required a password to be given :))

Thanks for testing. Yes, forgot to mention that password. I will add.

leivos-unikie:
> Thanks for the improvements! Now this works great. I tested running both pre-merge-tests and bat-tests in Ghaf from ghaf-host and then checked the results in Chrome.
> 
> > Or is it replacing your previously set user with the testuser?
> 
> It won't replace the user. If the username/password are different the tests just won't work.
> 
> (As a note about the documentation: `python ../lib/send_report.py` also required a password to be given :))

Thanks for testing. Yes, forgot to mention that password. I will add.

PR #221, opened 2025-01-29
Keep order of performance results

azbeleva:
current view in correct order:
image

### DIFF
--- a/Robot-Framework/lib/PerformanceDataProcessing.py
+++ b/Robot-Framework/lib/PerformanceDataProcessing.py
@@ -824,7 +824,7 @@ class PerformanceDataProcessing:
         tests = ['cpu_1thread', 'memory_read_1thread', 'memory_write_1thread', 'cpu', 'memory_read', 'memory_write']
         data = {test: {} for test in tests}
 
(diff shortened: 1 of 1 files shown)


Robot-Framework/lib/PerformanceDataProcessing.py
```diff
@@ -854,14 +854,16 @@ def read_vms_data_csv_and_plot(self, test_name, vms_dict):
                             'values': [build[1] for build in build_data],
                             'threads': threads
                         }
-                        all_builds[test].update([build[0] for build in build_data])
+                        for build in [build[0] for build in build_data]:
```
leivos-unikie, line 857: The variable naming makes this a bit hard to follow. Is it actually iterating here over commit hashes (or older build ids)?

Anyway it seems to work.

azbeleva: yeah it is actually commits, but I would say that every record (every row) is one test run (one build) so that's why I kept 'builds' naming, commits can repeat and they will get a counter (-1,-2, etc)

leivos-unikie (approved):

PR #226, opened 2025-01-31
fix asyncclick

avnik:


leivos-unikie (approved):

milva-unikie (approved):
Works also for me with older Nixos version (24.05)

PR #228, opened 2025-02-04
Adapt to ghaf pr915

leivos-unikie:
Continuing from https://github.com/tiiuae/ci-test-automation/pull/222 (which was later reverted by https://github.com/tiiuae/ci-test-automation/pull/227) to adapt ci-test-automation to new network settings:

X1: ethernet-USB adapter connects to net-vm, not anymore to ghaf-host
Orin-AGX: ethernet interface still on ghaf-host (same as before)
Orin-NX: ethernet-USB adapter interface still on ghaf-host and ethernet interface on net-vm (same as before)


milva-unikie (approved):
Tested bat-tests on X1. Nixos version test fails, others pass. Let's fix that later.

PR #230, opened 2025-02-05
Change time tolerance to be 30s.

msaarine:
Allow 30s tolerance between timestamps instead of 10s.


msaarine:
This was not good, created a new PR with different changes.

PR #231, opened 2025-02-05
Change timesync tolerance to be 30s instead of 10s.

msaarine:
Timesync -test: Changed tolerance to be 30s instead of default one used in python -script (10s)


leivos-unikie:
I was also looking those timesync tests more closely. I think the naming of ${change_time} variable is confusing. How about replacing those with ${original_time}?

leivos-unikie (approved):

PR #232, opened 2025-02-05
Remove check against hard coded limits in iperf result analysis

leivos-unikie:
There was an extra layer of harcoded check for iperf tests in keyword Get Throughput Values in addition of applying the adaptative result analysis tooling of PerformanceDataProcessing. Pass/fail criteria of that extra check was based on single hard coded limit for all targets. 

However results show that different targets have different baselines. The existing general result analysis tooling is able to adapt to different baselines on different targets and can detect any significant deviation from baseline. Extra check is not needed.

Added some extra logging in case measurement result is empty.


msaarine (approved):
Looked ok for me.

PR #235, opened 2025-02-06
Fix error in reset scaling keyword

leivos-unikie:


milva-unikie (approved):
Good find, thanks for fixing it!

PR #238, opened 2025-02-11
Add iterations to Connect keyword

leivos-unikie:
After ethernet interface was passed through to net-vm on lenovo-x1 iterative check of login to net-vm was first added to boot tests of lenovo-x1. Failures after gui reboot tests showed that iterations have to be added to the Connect keyword instead of using them only in the boot test.

The default number of iterations in Connect is 1 (as previous) but in boot tests and gui reboot 10 iterations are given.

Also Orin-NX needs to have iterative Connect keyword at boot and the output target has to be net-vm as with lenovo-x1. Both lenovo-x1 and orin-nx expose ghaf-host for about 10 sec time window at boot. During that time window ssh login attempts end up to ghaf-host.

- Modified also `Connect to ghaf host` to connect via net-vm on Orin-NX (as with lenovo-x1).

- Removed orin-nx tag from those test cases which include stopping net-vm. 

- `Connect to ghaf host` in optee suite setup instead of plain `Connect`

- Switching between connections didn't work anymore in `Verify NetVM PCI device passthrough`. Fixed that.

### DIFF
--- a/Robot-Framework/resources/common_keywords.resource
+++ b/Robot-Framework/resources/common_keywords.resource
@@ -82,19 +82,6 @@ Check If Device Is Up
         END
     END
 
-Check if netvm has started on Lenovo-X1
-    [Arguments]    ${iterations}=10
-    FOR    ${i}    IN RANGE    ${iterations}
-        ${pass_status}  ${connection}  Run Keyword And Ignore Error    Connect    target_output=ghaf@net-vm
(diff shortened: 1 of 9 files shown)


msaarine (approved):
No other comments but is '${iterations}=5'  enough?

milva-unikie (approved):

PR #240, opened 2025-02-14
Prevent fail in connect iteration to fail the keyword

leivos-unikie:
It was noticed that the whole Connect keyword can fail if Login of single iteration fails. With this change it will be avoided. 

In the beginning of boot ethernet interface is on ghaf-host but gets then passed through to net-vm at some point. If the switch of ethernet interface from ghaf-host to net-vm happens between Open Connection and Login keywords it is possible that the opened connection is actually lost in between and thus Login fails. This can randomly fail the Login keyword of the iteration at boot.


leivos-unikie:
Passed boot test on lenovo-x1
https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/473/artifact/Robot-Framework/test-suites/lenovo-x1ANDrelayboot/report.html

msaarine (approved):
Seems ok.

PR #247, opened 2025-02-18
Add systemctl status test for Lenovo

azbeleva:


milva-unikie (approved):
Confirmed locally that this works

PR #268, opened 2025-03-13
Add dell-7330 tag

azbeleva:


milva-unikie (approved):
Looks good to me if it passes in tests

PR #277, opened 2025-03-24
Change plot_dir to relative path

leivos-unikie:
In case of using non-default output directory the path to performance plots was set absolute. If report.html is opened from the location it was originally created it can show the links to performance plots but if the result folder is moved to another location the links to plots are broken in report.html.

This commit sets relative path to performance plots also for non-default output directory as it is the case for default output directory.


leivos-unikie:
Tested this locally that report.html has performance plots visible also when non-default output dir is used and output dir is copied to different location after test run.

msaarine (commented):
Didn't try this myself but seems quite straightforward change.

msaarine (approved):

PR #279, opened 2025-03-25
Exclude welcome check for Lenovo & Dell.

msaarine:
Excluding 'Welcome to NixOS' waiting from Lenovo & Dell, it is enough to check if process status and state are as expected.

Tests on Orin-AGX: [782](https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/782/robot/report/log.html)


milva-unikie (approved):
Tested on Lenovo-X1, fixes the issue I was having

azbeleva (approved):

PR #281, opened 2025-03-26
Support Cosmic

milva-unikie:
Add support for running Lenovo-x1 tests on Cosmic. There should be no major changes when Labwc is in use. Compositor is detected automatically. 

### List of changes for Cosmic

* Use `Tab and enter` to navigate app menu & power menu instead of image recognition and clicking. Those menus don't recognize teleported mouse and clicking does not work.
    * I tried using wlrctl tool instead of ydotool, but Cosmic does not yet support zwlr_virtual_pointer_manager_v1.
* New keywords `Log in, unlock and verify` and `Log out and verify` that take care of the whole login/logout operation.
* `Log out via GUI` uses Super+Shift+Escape keyboard shortcut to log out.  It should be more robust than opening power menu and clicking from there. Power menu log out is tested in GUI tests. 
* `Verify logout` renamed to `Check if logged out`. It will now return boolean instead of setting global variable. `systemctl --user is-active ghaf-session.target` can be used to verify if user is logged in or not.
* No need to reset scaling anymore after logging in, with cosmic it is the same with both types of Gen11 laptops.
* Do not disturb needs to be turned on while running GUI tests. The X on notification is the same that the X on app window and the image recognition can't separate those.
* Gui-vm apps need display set to be able to launch, new argument for that added to `Start XDG application`
* For some Cosmic apps there are other processes for nearly same name running. New argument added for `Find pid name` to be able to grep the correct process.
* Removed unnecessary `Start ydotoold` commands. It needs to be restarted every time after Close all Connections or reboot, other than that it stays active.
* New teardown for GUI app tests. All connections are closed between test cases to limit the amount of active connections. It makes running the tests a bit slower but a lot more reliable.

### Test runs

* [Main pipeline run with Labwc](https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/ghaf-main-pipeline/128/)
  * Dell did not boot, but separate test run can be found below
* [Lenovo-X1 Labwc - all tests](https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/1211/) (performance tests excluded)
* [Lenovo-X1 Cosmic - all tests](https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/1211/) (performance tests excluded)
* [Dell-7330 all tests](https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/1215/) (performance tests excluded)

### DIFF
--- a/Robot-Framework/resources/gui_keywords.resource
+++ b/Robot-Framework/resources/gui_keywords.resource
@@ -10,75 +10,139 @@ Library             Collections
 
 *** Variables ***
 
-${START_MENU}         ./launcher.png
-${LOGGED_IN_STATUS}   ${True}
-${DISABLE_LOGOUT}     ${EMPTY}
-${GUI_TEMP_DIR}       ${OUTPUT_DIR}/outputs/gui-temp/
+${APP_MENU_LAUNCHER}   ./launcher.png
+${LOCK_ICON}           ./lock.png
+${DISABLE_LOGOUT}      ${EMPTY}
+${GUI_TEMP_DIR}        ${OUTPUT_DIR}/outputs/gui-temp/
 
 *** Keywords ***
 
-Log in via GUI
-    [Documentation]   Login and verify that task bar is available
-    [Arguments]       ${stop_swayidle}=True
-    Switch Connection    ${CONNECTION}
-    Verify logout
-    IF  ${LOGGED_IN_STATUS}
-        Log To Console          Already logged in. Skipping login.
+Log in, unlock and verify
+    [Documentation]   Open desktop by logging in or unlocking. Verify that desktop is available.
+    [Arguments]       ${stop_swayidle}=True  ${enable_dnd}=False
+    Start ydotoold
+    ${logout_status}    Check if logged out   1
+    IF  ${logout_status}
+        Log in via GUI
+        Wait for Ghaf session activation
     ELSE
-        Log To Console          Logged out state detected. Logging in.
-        Start ydotoold
-        Log To Console          Typing username and password to login
-        Type string and press enter  ${USER_LOGIN}
-        Type string and press enter  ${USER_PASSWORD}
+        ${lock}       Check if locked
+        IF  ${lock}   Unlock
     END
     Try to reset scaling
-    Verify login
-    IF  ${stop_swayidle}
-        Stop swayidle
-    END
+    Verify desktop availability
+    IF  ${stop_swayidle}   Stop swayidle
+    IF  ${enable_dnd}      Set do not disturb state   true
 
-Log out
-    [Documentation]   Log out and optionally verify that desktop is not available
+Log out and verify
+    [Documentation]   Log out. Verify that session was logged out.
+    [Arguments]       ${disable_dnd}=True
     # Allow disabling logout in case of running test automation locally from ghaf-host.
     # This prevents terminal from being shutdown and allows test run to finish.
     IF  $DISABLE_LOGOUT == 'true'
         Log To Console    Log out disabled. Skipping log out procedure.
         RETURN
     END
+    IF  ${disable_dnd}   Set do not disturb state   false
     Start ydotoold
-    Get icon           ghaf-artwork  power.svg  crop=0  background=black
-    Locate and click   icon.png  0.95  5
-    Get icon           ghaf-artwork  logout.svg  crop=0  background=black
-    Locate and click   icon.png  0.95  5
+    ${logout_status}      Check if logged out    1
+    IF  not ${logout_status}
+        Log out via GUI
+        ${logout_status}  Check if logged out
+        IF  not ${logout_status}    FAIL  Failed to log out.
+    END
+    Stop ydotoold
+
+Log in via GUI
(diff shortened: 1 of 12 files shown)


Robot-Framework/test-suites/gui-tests/gui_apps.robot
```diff
@@ -9,130 +9,145 @@ Resource            ../../config/variables.robot
...
-Start and close chrome via GUI on LenovoX1
-    [Documentation]   Start Chrome via GUI test automation and verify related process started
-    ...               Close Chrome via GUI test automation and verify related process stopped
-    [Tags]            SP-T41   lenovo-x1
-    Get icon   app  google-chrome.svg  crop=30
-    Start app via GUI on LenovoX1   ${CHROME_VM}  chrome
+Start and close Google Chrome via GUI on LenovoX1
+    [Documentation]   Start Google Chrome via GUI test automation and verify related process started
+    ...               Close Google Chrome via GUI test automation and verify related process stopped
+    [Tags]            SP-T41-2  #lenovo-x1 feature is broken (border missing)
```
azbeleva, line 27: should we use "skip" and define jira key of the issue?

Robot-Framework/resources/gui_keywords.resource
```diff
@@ -10,72 +10,99 @@ Library             Collections
...
-    IF  ${LOGGED_IN_STATUS}
-        Log To Console          Already logged in. Skipping login.
+Log in, unlock and verify
+    [Documentation]   Open desktop by logging in or unlocking. Verify that desktop is available.
+    Start ydotoold
+    ${logout_status}    Check if logged out    1
+    IF  ${logout_status}
+        Log in via GUI
+        # Give desktop time to load, needed especially for Dell and when external monitor is connected.
+        Sleep  15
```
azbeleva, line 27: are not there any other ways to check that screen is ready? 'sleep' is kinda okay but not the best solution in common

msaarine (commented):
Seemed ok for me.

Robot-Framework/resources/gui_keywords.resource
leivos-unikie: > are not there any other ways to check that screen is ready? 'sleep' is kinda okay but not the best solution in common

How about iterating and trying to take grim screenshots like in "Verify logout" keyword (it fails in logged out state). When the screenshot succeeds then could iterate if the screenshot has launcher icon visible...

milva-unikie: In Cosmic we don't have grim, instead there is cosmic-screenshot. The issue with cosmic-screenshot is that if display environment variable is not available, it gets stuck and eventually core dumps. It can not be used to confirm the availability same way that grim was.

With just Lenovo-X1 sleep is not needed. When external display is connected to Lenovo-X1 5 seconds of sleep is fine. But when testing with Dell 10 seconds was not always enough. Just increased it to 15 so that it is not blocking everything. If login fails almost non of the tests are able to run.

There probably is some better way, I will look into it.

milva-unikie: I talked with @kajusnau and I think I have a solution. Will need to confirm it with Dell next week.

Robot-Framework/test-suites/gui-tests/gui_apps.robot
milva-unikie: Skip added, there are no Jira issues yet because the PR is still open

Robot-Framework/resources/gui_keywords.resource
milva-unikie: Replaced the sleep by with checking if ghaf-session is active. `Locate image on screen` is also improved so that it won't hang even if display environment variable would not be available.

Robot-Framework/test-suites/gui-tests/gui_power_options.robot
```diff
@@ -25,9 +25,12 @@ GUI Suspend and wake up
     [Tags]            lenovo-x1   SP-T208-2
     Start power measurement       ${BUILD_ID}   timeout=180
     Set start timestamp
-    Connect to netvm
-    Connect to VM                 ${GUI_VM}
```
leivos-unikie, line 29: I think here you should still have
    Connect to netvm
    Connect to VM                 ${GUI_VM}
    
otherwise current ssh connection is left on measurement agent (see: Start power measurement) and robot attempts to run the following ssh library commands there (if the setup has RPi).

leivos-unikie: It works on dev though, because it does not have power measurement tooling.

milva-unikie: Good find. Connecting got accidentally deleted when I created a separate test setup that takes care of the connecting for other test cases. Added it back and rebased to also include same for the suspension test.

leivos-unikie (approved):

PR #288, opened 2025-04-01
Fix path for power plot

leivos-unikie:
report.html file link to power plot has been broken after changing to path variables. Fix the problems.

### DIFF
--- a/Robot-Framework/lib/ParsePowerData.py
+++ b/Robot-Framework/lib/ParsePowerData.py
@@ -5,12 +5,21 @@ import pandas as pd
 import logging
 import matplotlib.pyplot as plt
 import csv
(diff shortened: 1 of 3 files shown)


leivos-unikie:
Tested this on dev jenkins and got report.html to show the power plot
https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/891/artifact/Robot-Framework/test-suites/lenovo-x1ANDSP-T208-2/report.html

Tested that it works even locally with other than default --outputdir (reading dummy measurement values from RPi setup connected to lenovo-x1 on prod).

azbeleva (approved):

PR #289, opened 2025-04-02
Measure dynamic increase and decrease of memory

leivos-unikie:
Ballooning monitor should allocate more memory for a VM when its available memory approaches 0. 

This test case is first consuming and then releasing a lot of memory while simultaneously logging total, used and available available memory of the VM. Test keyword is generic, can be used for any VM but suitable parameters have to be checked if applying to other VMs than those now applied.

All VMs don't have memory ballooning enabled. Other VMs can have different maximum limit for inflating total memory.

https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/979/

***
Checked that currently on ghaf side the default limit for memory ballooning has been set to 2 x vm initial memory:
(modules/microvm/appvm.nix)
```
            balloonRatio = mkOption {
              description = ''
                Amount of dynamic RAM for this AppVM as a multiple of ramMb
              '';
              type = types.number;
              default = 2;
            };
```

E.g. for chrome-vm the initial memory is set as
modules/reference/appvms/chromium.nix
`ramMb = 6144;`

If these parameters are tuned (decreased) on ghaf side ballooning test parameters need to be tuned respectively.

### DIFF
--- a/Robot-Framework/lib/ParseMemoryLogData.py
+++ b/Robot-Framework/lib/ParseMemoryLogData.py
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2022-2025 Technology Innovation Institute (TII)
+# SPDX-License-Identifier: Apache-2.0
+
+import pandas as pd
+import logging
+import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
+
+
+class ParseMemoryLogData:
+
+    def __init__(self, perf_data_dir):
+        self.perf_data_dir = perf_data_dir
+
+    def generate_graph(self, plot_dir, id):
+        data = pd.read_csv(self.perf_data_dir + "ballooning_" + id + ".csv")
+        start_time = 0
+        end_time = int(data['time'].values[data.index.max()])
+        step = int((end_time - start_time) / 20)
+        if step < 1:
+            step = 1
+        plt.figure(figsize=(20, 10))
+        plt.set_loglevel('WARNING')
+        plt.ticklabel_format(axis='y', style='plain')
+        plt.yticks(fontsize=14)
+        plt.plot(data['time'], data['total_mem'], marker='o', linestyle='-', color='b', label='total_mem')
+        plt.plot(data['time'], data['used_mem'], marker='o', linestyle='-', color='g', label='used_mem')
+        plt.plot(data['time'], data['available_mem'], marker='o', linestyle='-', color='r', label='avail_mem')
+        plt.title('Memory ballooning', loc='center', fontweight="bold", fontsize=16)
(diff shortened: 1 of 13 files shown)


Robot-Framework/lib/ParseMemoryLogData.py
```diff
@@ -0,0 +1,45 @@
...
+        plt.plot(data['time'], data['available_mem'], marker='o', linestyle='-', color='r', label='avail_mem')
+        plt.title('Memory ballooning', loc='center', fontweight="bold", fontsize=16)
+        plt.ylabel('MegaBytes', fontsize=16)
+        plt.grid(True)
+        plt.xlabel('Time (s)', fontsize=16)
+        # plt.xticks(data['time'], rotation=45, fontsize=14)
+        plt.legend(loc='upper left', fontsize=20)
+
+        # Set maximum for tick number
+        # ax.xaxis.set_major_locator(plt.MaxNLocator(30))
```
azbeleva, line 37: is this commented code here on purpose?

leivos-unikie: no, I will detete those

Robot-Framework/lib/ParseMemoryLogData.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: 2022-2024 Technology Innovation Institute (TII)
```
azbeleva, line 1: 2025?

Robot-Framework/test-suites/performance-tests/consume_memory
```diff
@@ -0,0 +1,30 @@
...
+done
+
+# Fill the memory by copying the source file
+for index in $(seq ${write_iterations}); do
+    cp /dev/shm/test/source /dev/shm/test/file${index}
+    if [[ $(cat ${status_file}) == "stage1" ]]
+    then
+        break
+    fi
+    # sleep 0.001
```
azbeleva, line 26: commented

azbeleva (approved):

Robot-Framework/test-suites/performance-tests/consume_memory
leivos-unikie: Removed the commented delay. Delay is not necessary with current memory/cpu speeds when using the defined source file size.

Robot-Framework/lib/ParseMemoryLogData.py
leivos-unikie: Changed to 2025. Went also through all other files having still "2022-2024" and updated them to "2022-2025".

PR #292, opened 2025-04-07
Fix tx/rx typo.

msaarine:
Noticed that tx/rx are mixed up when creating message


azbeleva (approved):

leivos-unikie (approved):
I am just wondering what is that last line in network.robot?

Could you remove that?

msaarine:
> I am just wondering what is that last line in network.robot?
> 
> Could you remove that?

Removed.

PR #310, opened 2025-05-02
Re-establish 'normal' connection after connecting to RPI in PROD

msaarine:
Reconnecting device in case of power measurement usage in PROD.
The test worked fine in DEV since the RPI connection is not  used there.

Test result(PROD): https://ghaf-jenkins-controller-prod.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/180/ 
Test result (DEV): https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/1224/

### DIFF
--- a/Robot-Framework/test-suites/suspension-test/suspension.robot
+++ b/Robot-Framework/test-suites/suspension-test/suspension.robot
@@ -32,6 +32,8 @@ Automatic suspension
     Check screen brightness  ${max_brightness}
 
     Start power measurement       ${BUILD_ID}   timeout=180
+    Connect
+    Connect to VM    ${GUI_VM}  ${USER_LOGIN}  ${USER_PASSWORD}
     Set start timestamp
 
     Wait     240


milva-unikie (approved):

Robot-Framework/test-suites/suspension-test/suspension.robot
```diff
@@ -32,6 +32,7 @@ Automatic suspension
     Check screen brightness  ${max_brightness}
 
     Start power measurement       ${BUILD_ID}   timeout=180
+    Re-establish connection if RPI was connected
```
leivos-unikie, line 35: Or would it be enough just to do here:

```
Connect
Connect to VM    ${GUI_VM}  ${USER_LOGIN}  ${USER_PASSWORD}
```

Robot-Framework/test-suites/suspension-test/suspension.robot
```diff
@@ -32,6 +32,9 @@ Automatic suspension
     Check screen brightness  ${max_brightness}
 
     Start power measurement       ${BUILD_ID}   timeout=180
+    # Re-establish connection (if RPI is connected, it will cause disconnection of original connection.
```
leivos-unikie, line 35: I think here we replace the original setup connection anyways, whether or not RPi was available.

For RPi keywords I made previously decision not to use Switch Connection keyword in those at all. It caused confusion when power measurement was implemented in multiple test cases (having Close All Connections in between).

leivos-unikie (approved):

Robot-Framework/test-suites/suspension-test/suspension.robot
msaarine: Yes, seems like that does the trick (I tried several connection combinations but missed this)

Robot-Framework/test-suites/suspension-test/suspension.robot
msaarine: OK, removed the comment. Now the connection is set up always.

PR #311, opened 2025-05-05
Give enough time for the power measurement script

leivos-unikie:
The timeout used to launch the power measurement script on the measurement agent was only 180 sec but according to the test report logs the Automatic suspension test lasts around 20 min 30 sec. Increase the timeout well beyond the duration of the test case, to 1500 sec.

Then we will see power plot over the full automated suspension and wake up, not just 3 min from the beginning when suspension hasn't yet kicked in.


msaarine (commented):
I wondered last week if there was something wrong with that plot, this explains it totally.

msaarine (approved):

PR #312, opened 2025-05-05
Fix Orin-NX connection issues.

msaarine:
Orin-NX performance tests had failed to execute, the setup keyword had failed.
When checking this, it was noticed that connection to 'net-vm' was not established but connected tartget was still 'ghaf@ghaf-host'
There was just one  'open connection'  trial set as default for loop, that was increased to be '3' and now the connection seem to take place.

Failed run: [perf-5930](https://ghaf-jenkins-controller-prod.northeurope.cloudapp.azure.com/job/ghaf-hw-test/5930/artifact/Robot-Framework/test-suites/performance/log.html)

With fixes [dev-1231](https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/1231/robot/report/log.html)

### DIFF
--- a/Robot-Framework/resources/ssh_keywords.resource
+++ b/Robot-Framework/resources/ssh_keywords.resource
@@ -58,7 +58,7 @@ Login with timeout
 
 Connect
     [Documentation]   Set up the SSH connection to the device
-    [Arguments]       ${IP}=${DEVICE_IP_ADDRESS}    ${PORT}=22    ${target_output}=None   ${iterations}=1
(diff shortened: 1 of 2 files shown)


leivos-unikie:
After boot time test the `Connect to ghaf host` keyword of the following performance test comes too quickly. There is still the risk of getting stuck if the switch of eth interface from ghaf-host to net-vm happens between `Open Connection` and `Login with timeout` keywords. 

This was seen previously in boot tests and consequently 30 sec sleep was added there before ssh connection check. Same 30 sec sleep could be applied in the end of `Measure Orin Soft Boot Time` / `Measure Orin Hard Boot Time` in case of NX target.

leivos-unikie (approved):

PR #321, opened 2025-05-12
Give more time for timesync

leivos-unikie:
Sometimes it requires more than currently set 3 sec. Let's set it to 20 sec.

Example of failing to get the time synchronized in 3 sec after switching timesync service on:
https://ghaf-jenkins-controller-prod.northeurope.cloudapp.azure.com/job/ghaf-hw-test/6186/artifact/Robot-Framework/test-suites/bat/log.html#s1-s1-s10-t1


msaarine (approved):
I was going to do that same change after checking the results, but you was faster.

PR #324, opened 2025-05-20
Add Falcon AI test

azbeleva:
test on Dell:
https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/1429/
test on Lenovo:
https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/1432/

### DIFF
--- a/Robot-Framework/resources/common_keywords.resource
+++ b/Robot-Framework/resources/common_keywords.resource
@@ -4,6 +4,7 @@
 *** Settings ***
 Resource            ../config/variables.robot
 Resource            ../resources/gui_keywords.resource
+Library             ../lib/output_parser.py
 
 
 *** Keywords ***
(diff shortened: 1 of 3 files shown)


Robot-Framework/resources/common_keywords.resource
```diff
@@ -90,3 +91,8 @@ Log versions
     Log to console      Ghaf version: ${ghaf_version}
     ${nixos_version}    Execute Command   nixos-version
     Log to console      Nixos version: ${nixos_version}
+
+Get mako path
+    ${output}            Execute Command     ls /nix/store | grep mako | grep -v .drv
+    ${result}            Extract mako path   ${output}
+    Set Global Variable  ${MAKO_PATH}    ${result}
```
milva-unikie, line 98: Why set a global variable instead of simply returning the path?

azbeleva: I use it in different test set, several time, so I can return the variable and then assign it to some site variables after that, so this just give us less lines of code

milva-unikie: That's fair :)

milva-unikie (approved):

PR #375, opened 2025-07-02
Add Press Key(s) keyword

milva-unikie:
Add `Press Key(s)` keyword to call when using `ydotool key` command. It generates the command from the key names and runs it.
[
Lenovo-X1 testrun](https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/2354/) (the two tests that failed also failed in nightly pipeline)

Advantages:

- Makes the code easier to read -> `Press Key(s) LEFTMETA+LEFTSHIFT+ESC` instead of `Execute Command ydotool key 125:1 42:1 1:1 1:0 42:0 125:0 sudo=True sudo_password=${PASSWORD}`
- Removes the need to create new keywords for keys that are only used a few times. We can still have keywords for commonly used combinations. For example I removed `Increase brightness` &  `Decrease brightness` but left the `Switch keyboard layout` keyword.
- Easier error handling, already added a check for the return code. We can also catch errors if we want. Right now errors returned by ydotool are just ignored and can cause confusion when the test later fails.

### DIFF
--- a/Robot-Framework/resources/gui_keywords.resource
+++ b/Robot-Framework/resources/gui_keywords.resource
@@ -41,7 +41,7 @@ Log in, unlock and verify
     IF   $status == 'FAIL'
         Log To Console    There was maybe a bug with the cosmic-greeter, let's try logging in again
         Log To Console    Pressing Tab to activate the password field
-        Execute Command    ydotool key 15:1 15:0  sudo=True  sudo_password=${PASSWORD}
+        Press Key(s)      TAB
         Log in via GUI
         Wait for Ghaf session activation
         Verify desktop availability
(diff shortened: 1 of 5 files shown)


azbeleva (approved):

Robot-Framework/resources/gui_keywords.resource
```diff
@@ -85,9 +85,9 @@ Log out via GUI
         Log To Console    Log out disabled. Skipping log out procedure.
         RETURN
     END
-    Log To Console    Logging out by pressing Super+Shift+Escape 
-    Execute Command   ydotool key 125:1 42:1 1:1 1:0 42:0 125:0  sudo=True  sudo_password=${PASSWORD}
-    Tab and enter     tabs=2
+        Log To Console    Logging out by pressing Super+Shift+Escape 
+        Press Key(s)      LEFTMETA+LEFTSHIFT+ESC
+        Tab and enter     tabs=2
```
azbeleva, line 90: extra tabs?

milva-unikie: Thanks, fixed

PR #378, opened 2025-07-02
Generalize keyword for switch to vm

leivos-unikie:
Create general keyword which can be used in switching to any vm like Switch to gui-vm as ghaf/user.

Replace Switch to gui-vm as ghaf/user keywords with

`Switch to vm    gui-vm`
and 
`Switch to vm    gui-vm  user=${USER_LOGIN}`

**NOTES:** 

After using `Close All Connections` keyword previously automatically created connection index variables still exist and have some value but the connections related to the indices are lost. So even though `Check variable availability` passes it doesn't ensure that the connection is still active. `Try to switch to active gui-vm connection` is still very important in handling the situations after `Close All Connections`.

I read from Robot Framework documentation that `Switch Connection` actually returns the index of the previous connection not index of the new connection switched to. Thus renamed the variable:
`${connection_status}  ${new_connection}   Run Keyword And Ignore Error    Switch Connection   ${ssh_connection}`
-->
`${connection_status}  ${previous_connection}   Run Keyword And Ignore Error    Switch Connection   ${ssh_index} `

**Regression test run on Dell**
https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/2403/

Testing vms after adding 
`Library             ../lib/output_parser.py` 
back to `ssh_keywords.resource`
https://ghaf-jenkins-controller-dev.northeurope.cloudapp.azure.com/job/tests/job/x-ghaf-hw-test/2409/

### DIFF
--- a/Robot-Framework/resources/ssh_keywords.resource
+++ b/Robot-Framework/resources/ssh_keywords.resource
@@ -9,11 +9,8 @@ Library             String
 Library             Process
 Library             String
 Library             ../lib/output_parser.py
+Library             ../lib/helper_functions.py
 
-*** Variables ***
-
-${GUI_VM_GHAF_SSH}     ${EMPTY}
-${GUI_VM_USER_SSH}     ${EMPTY}
 
 *** Keywords ***
 
@@ -122,7 +119,7 @@ Connect to netvm
 Connect to VM
     [Documentation]      Connect to any VM or ghaf-host over internal virtual network
     [Arguments]          ${vm_name}    ${user}=${LOGIN}   ${pw}=${PASSWORD}   ${timeout}=60
-    Log To Console       Connecting to ${vm_name} as ${user}...   no_newline=true
+    Log                  Connecting to ${vm_name} as ${user}...   console=True
     Check if ssh is ready on vm        ${vm_name}   ${timeout}
     ${failed_connection}  Set Variable  True
     ${start_time}  Get Time	epoch
@@ -148,7 +145,6 @@ Connect to VM
         BREAK
     END
     IF  ${failed_connection}    FAIL  Couldn't connect ${vm_name}
-
     # Logging in to vm
     ${logged_in}  Set Variable  False
     FOR    ${i}    IN RANGE     5
@@ -159,7 +155,6 @@ Connect to VM
     END
     IF  not ${logged_in}  FAIL  Could not login to ${vm_name} as ${user}.
     Log To Console  Connected and logged in.
-    
     # If connected and logged in successfully
     RETURN  ${connection}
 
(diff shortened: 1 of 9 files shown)


Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -12,8 +12,8 @@ Library             ../lib/output_parser.py
 
 *** Variables ***
 
-${GUI_VM_GHAF_SSH}     ${EMPTY}
-${GUI_VM_USER_SSH}     ${EMPTY}
+${GUI-VM_GHAF_SSH}     ${EMPTY}
+${GUI-VM_USER_SSH}     ${EMPTY}
```
milva-unikie, line 16: These variables can be removed

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -692,32 +692,28 @@ Create test user
...
-        Set Global Variable    ${GUI_VM_USER_SSH}
+    IF  ${connection_exists}
+        ${status}   Try to switch to active vm connection   ${${ssh_connection}}   ${hostname}   ${user}
+        IF   ${status}
+            Log    Switched to ${hostname} as ${user}   console=True
+            RETURN
+        END
     END
+    ${${ssh_connection}}      Connect to VM       ${hostname}
+    Set Global Variable       ${${ssh_connection}}
```
milva-unikie, line 713: Is this `${${ssh_connection}}` intentional?

leivos-unikie: Yes, it's intentional. We have to get the variable name from variable ${ssh_connection} and compose a new variable with that, not just set ${ssh_connection} a global variable.

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -8,12 +8,8 @@ Library             BuiltIn
 Library             String
 Library             Process
 Library             String
-Library             ../lib/output_parser.py
+Library             ../lib/helper_functions.py
```
milva-unikie, line 11: `../lib/output_parser.py` should be still be here

leivos-unikie: ah, ok. I guess that caused `VM Suite Setup` to fail

So ../lib/output_parser.py should be inherited via ssh_keywords.resource to vm.robot

milva-unikie: > So ../lib/output_parser.py should be inherited via ssh_keywords.resource to vm.robot

Yes, though it does not hurt if it is also defined in vm.robot. `ssh_keywords.resource` needs it included anyways.

PR #405, opened 2025-08-07
Refactor setups and teardowns

azbeleva:
Changes:
1. Now there is common __init__ file for ALL test suites, where "global" setup is defined. Here we initialize variables. This will require to use these variable in all other test suiltes' init files, especially use device control library.
2. Suspension test: after waking up od the device - check that lock screen appeared. In case is test fails, reboot laptop.
3. Replace "Log To Console" with "Log" + console=True for comments with variables, to be able to see them in reports, not only in logs.

Example of the failed suspension test with reboot in the test teardown: https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/408/artifact/Robot-Framework/test-suites/SP-T162/log.html
Here are 3 levels of setups/teardowns:
image

### DIFF
--- a/Robot-Framework/resources/common_keywords.resource
+++ b/Robot-Framework/resources/common_keywords.resource
@@ -42,11 +42,11 @@ Verify Systemctl status
         Set Test Variable   ${failed_units}
 
         IF  '${status}' not in ['running', 'starting']
-            Log To Console   Systemctl status is ${status}
+            Log     Systemctl status is ${status}  console=True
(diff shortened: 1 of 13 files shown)


milva-unikie (approved):
In suspension teardown there is only "Reboot Laptop" step but no time for the laptop to actually boot. I guess the wait time in "Connect to device" should be long enough for the next test to get the connection?

milva-unikie:
> `the wait time in "Connect to device" should be long enough for the next test to get the connection?` yes, in the linked testrun after reboot connection in suite teardown took ~1min

Oh true, I missed that

PR #413, opened 2025-08-14
Use text recognition in login and other small changes

milva-unikie:
* Use text "Applications" to confirm login instead of the launcher icon. The icon is sometimes missing or hidden.
* Use text "Password" to confirm if lock is active instead of the lock icon. The same lock icon is also used in the power menu so it could cause issues.
* => Icons are now only needed for gui tests.
* Change to gray wallpaper before login and back to default when tests are done. This improves image and text recognition.
* Remove unused `${stop_swayidle}` from `Prepare Test Environment` and from `Log in, unlock and verify`.
* Log device ID at the beginning of the tests.
* Log a proper message when image recognition failed as expected.
* Log gui-vm `systemctl status` before trying to take the first screenshot after login. This could provide useful information if the system gets stuck again.

[Lenovo-X1](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/589/)
[Orin-NX](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/603/) - device-id also works for Orins

### DIFF
--- a/Robot-Framework/resources/gui_keywords.resource
+++ b/Robot-Framework/resources/gui_keywords.resource
@@ -9,12 +9,12 @@ Documentation       Keywords for gui tests. Note: These keywords assume the conn
 Library             ../lib/GuiTesting.py    ${OUTPUT_DIR}/outputs/gui-temp/
 Library             Collections
 Library             DateTime
+Resource            ../resources/file_keywords.resource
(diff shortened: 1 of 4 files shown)


msaarine (approved):
Looked ok for me.

PR #417, opened 2025-08-15
Add skip for X1 'Automatic suspension' -test

msaarine:
In case of suspension test failure, set test as 'skipped' with related bug id.

[Test Run: ](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/637/)

### DIFF
--- a/Robot-Framework/test-suites/suspension-test/suspension.robot
+++ b/Robot-Framework/test-suites/suspension-test/suspension.robot
@@ -67,6 +67,7 @@ Test setup
 
 Test teardown
     Run Keyword If Test Failed    Reboot Laptop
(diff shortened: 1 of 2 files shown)


azbeleva (approved):

leivos-unikie (commented):
The test tends to fail particularly at `Check the screen state`, screen state not being on after ssh connection becoming available (Screen state: eDP-1 off). This seems to be different issue than https://jira.tii.ae/browse/SSRCSP-6986. 

Would the display state change if we just iterate `Check the screen state` some more. Or is the screen permanently off at this stage. Did you test / monitor? I was just thinking should there be a separate bug ticket "X1 display does not wake up from suspend" and link the Skip to that?

On my user experience X1 sometimes suspend (ssh connections drop) without screen shutting down. Sometimes screen shuts down when waiting longer and in wake up ssh becomes available but display does not wake up at all.

PR #445, opened 2025-09-16
Try to violate memory limit in ballooning test

leivos-unikie:
Changed ballooning test a lot. Now it tries to use memory as much as possible (even more than VM mem quota) by running the memory consume script simultaneously in three separate memory mounted directories.

Ballooning test checks that
- memory deflates at boot to defined (idle) level (vms start at fully ballooned state)
- memory manager allocates more memory to the vm when its available memory becomes small
- memory manager cannot allocate more memory than is defined by the VM mem quota + ballooning 
- memory deflates back to normal (idle) state after used memory is released

Test requires now 3 arguments:
- vm name
- mem quota (`vm.ramMb`) defined for the vm in ghaf repository
- max_inflate_ratio (currently default `balloonRatio`=2 which makes the max mem limit `3 x ramMb` -> max_inflate_ratio=3, see `mem = vm.ramMb * (vm.balloonRatio + 1);`) in `modules/microvm/appvm.nix`

Internal timeouts for inflate and deflate are now calculated from the amount of memory to be consumed.

Since the test is now trying to violate memory limits I had to pay more attention to units of memory.
- ghaf repository defines `vm.ramMb` in mebibytes (MiB)
- 1 MiB = 1024 x 1024 bits
- Use the same units in memory monitoring (`free --mebi`)

Tested locally on Lenovo-X1 multiple times.

Test run on Darter Pro
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/1283/

Testing revealed that with the current VM configuration Dell-7330 target does not survive memory stress test in a single VM. Ghaf seems to crash when memory of chrome-vm is consumed as much as possible: GUI freezes, ssh connections drop (however I didn't find OOM errors in journalctls). Dell-7330 has only 16GB of HW memory and it's memory is very overloaded already at idle state few minutes after boot:
```
[ghaf@ghaf-host:~]$ free --mega
               total        used        free      shared  buff/cache   available
Mem:           16494       16188         191       15079       15501         306
Swap:          12884        2965        9919
```

I monitored memory in ghaf-host during ballooning_chrome_vm test on Dell-7330 
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/1282/

The test gets stuck at 
`11:28:49  Used memory: 9424 / Total memory: 11909`

At that moment all memory resources (also 12GB swap) of ghaf-host are depleted:
```
Every 1.0s: free --mebi                                                                                                                                                            ghaf-host: Fri Sep 19 08:28:52 2025

               total        used        free      shared  buff/cache   available
Mem:           15730       15598         138       14559       14857         132
Swap:          12287       12279           8
```

Clearly ghaf VM configuration (mem quotas, balloonin ratios) should be tuned to more target specific (also fine tune among x86 targets). For now I exclude Dell-7330 from the ballooning tests. However, I improved Teardown so that even Dell-7330 can now survive test FAIL because of target device getting stuck (separate handling for SSHException) --> following tests can run.

### DIFF
--- a/Robot-Framework/test-suites/performance-tests/ballooning.robot
+++ b/Robot-Framework/test-suites/performance-tests/ballooning.robot
@@ -18,144 +18,150 @@ Test Timeout        10 minutes
 
 
 *** Variables ***
-${robot_status_file}      /tmp/ballooning_robot_status
-${script_status_file}     /tmp/ballooning_script_status
-${test_dir}               /tmp
+${test_dir}               /tmp/ballooning
+${script_status_file}     /tmp/ballooning/ballooning_script_status
 ${rebooted}               False
 
 
 *** Test Cases ***
 
 Test ballooning in chrome-vm
-    [Tags]                  ballooning_chrome_vm    lenovo-x1   darter-pro   dell-7330   SP-T255
-    Test ballooning in VM   vm=chrome-vm   max_init_memory=6500
+    [Tags]                  ballooning_chrome_vm    lenovo-x1   darter-pro   SP-T255
+    Test ballooning in VM   vm=chrome-vm   mem_quota=6144   max_inflate_ratio=3
 
 Test ballooning in business-vm
-    [Tags]                  ballooning_business_vm   lenovo-x1   darter-pro   dell-7330   SP-T256
-    Test ballooning in VM   vm=business-vm   max_init_memory=6500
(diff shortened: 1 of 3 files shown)


msaarine (approved):
Went through the PR-> Good findings & job in general!
Code seems to be consistent & OK.

The only this that caught my eye was echoing this: `'fillthememorywiththisdatafillthememorywiththisdata'` - is that intended or typo?

PR #457, opened 2025-09-26
Try to fix failures in connecting to ghaf-host

leivos-unikie:
Connection from net-vm to ghaf-host has failed often recently. 
Try fix the problem by decresíng creation of new connections by using the 'Switch to vm' keyword.


milva-unikie (commented):
Since the return value of `Connect to ghaf host` was removed can you also remove the CONNECTION variable
```
    ${CONNECTION}        Connect to ghaf host
    Set Global Variable  ${CONNECTION}
```
from `Connect to device` keyword? 

Other than that looks good.

msaarine (approved):

PR #761, opened 2026-04-10
Test givc policy update

leivos-unikie:
Add a test case `GIVC policy init on boot in policy-enabled VMs`
It checks that givc-policy-init
- is enabled
- is not failed
- journal does not show missing-file error
- configured policy destination files exist
- if stored_path exists, it matches the destination

For now ghaf givc policy is not mature for further tests but in the future more related test cases could be added as it develops.

Darter-Pro
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6743/

Lenovo-X1
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6744/

### DIFF
--- a/Robot-Framework/resources/service_keywords.resource
+++ b/Robot-Framework/resources/service_keywords.resource
@@ -7,6 +7,18 @@ Resource            ../resources/ssh_keywords.resource
 
 
 *** Keywords ***
+
+Service should not be failed
+    [Arguments]    ${service}
+    ${result}    Run Command    systemctl show -p Result --value ${service}    rc_match=skip
+    ${state}    Run Command    systemctl show -p ActiveState --value ${service}    rc_match=skip
+    IF    'failed' in '${result}' or 'failed' in '${state}'
+        ${service_journalctl}=    Run Command    journalctl -b -u ${service} --no-pager -n 200    rc_match=skip
+        Log                       ${service_journalctl}
+    END
+    Should Not Contain    ${result}    failed
+    Should Not Contain    ${state}    failed
+
 Verify service status
     [Documentation]   Check if service is running with given loop ${range}
     [Arguments]       ${range}=45  ${service}=${EMPTY}   ${expected_state}=active   ${expected_substate}=running  ${expected_rc}=0
--- a/Robot-Framework/test-suites/security-tests/givc_policy.robot
+++ b/Robot-Framework/test-suites/security-tests/givc_policy.robot
@@ -1,0 +1,113 @@
+# SPDX-FileCopyrightText: 2022-2026 Technology Innovation Institute (TII)
+# SPDX-License-Identifier: Apache-2.0
+
+*** Settings ***
+Documentation       Validate GIVC policy initialization and live policy update handling
+Test Tags           givc-policy  security  lenovo-x1  darter-pro  dell-7330
+
+Resource            ../../resources/common_keywords.resource
+Resource            ../../resources/ssh_keywords.resource
+
+Suite Setup         GIVC policy suite setup
+
+
+*** Variables ***
+${POLICY_BACKUP_DIR}    /tmp
+
+
+*** Test Cases ***
+
+GIVC policy init on boot in policy-enabled VMs
+    [Documentation]  Verify givc-policy-init exists and did not fail in VMs where policy client is enabled.
+    [Template]       Verify givc policy init in ${vm}
+    FOR    ${vm}    IN    @{VM_LIST_WITH_HOST}
+        ${vm}
+    END
+
+GIVC policy update triggers handler in policy-enabled VMs
+    [Documentation]  Modify an active policy file and verify its handler runs.
+    [Template]       Verify givc policy update handler in ${vm}
+    FOR    ${vm}    IN    @{VM_LIST_WITH_HOST}
+        ${vm}
+    END
+
+
+*** Keywords ***
+
+GIVC policy suite setup
+    @{VM_LIST_WITH_HOST}    Get VM list    with_host=True
+    Set Suite Variable      @{VM_LIST_WITH_HOST}
+
+Verify givc policy init in ${vm}
+    Switch to vm    ${vm}
+    ${rc}    Run Command    systemctl is-enabled givc-policy-init    return=rc    rc_match=skip
+    IF    ${rc}[0] != 0
+        SKIP    givc-policy-init not enabled in ${vm}
+    END
+    ${result}    Run Command    systemctl show -p Result --value givc-policy-init    rc_match=skip
+    ${state}     Run Command    systemctl show -p ActiveState --value givc-policy-init    rc_match=skip
+    Should Not Contain    ${result}    failed
+    Should Not Contain    ${state}     failed
+
+Verify givc policy update handler in ${vm}
+    Switch to vm    ${vm}
+    ${service}    ${path_unit}    ${policy_path}    Get GIVC policy handler details
+    IF    "${service}" == "${EMPTY}" or "${policy_path}" == "${EMPTY}"
+        SKIP    No givc policy handler detected in ${vm}
+    END
+
+    ${backup}    Set Variable    ${POLICY_BACKUP_DIR}/givc_policy_backup_${vm}
+    Run Command    cp ${policy_path} ${backup}    sudo=True
+
+    ${before}    Run Command    systemctl show -p ExecMainExitTimestampUSec --value ${service}    rc_match=skip
+    Run Command    sh -c "echo '# ci-test-automation policy update '$(date +%s)' ' | tee -a ${policy_path} >/dev/null"    sudo=True
+
+    Wait Until Keyword Succeeds    6x    5s    Policy handler should run    ${service}    ${before}
+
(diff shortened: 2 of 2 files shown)


Robot-Framework/test-suites/security-tests/givc_policy.robot
```diff
@@ -0,0 +1,122 @@
...
+    [Arguments]    ${policy_cfg}    ${policy_name}
+    ${store_path}    Evaluate    $policy_cfg["storePath"]
+    RETURN    ${store_path}/${policy_name}/policy.bin
+
+Verify givc policy init in ${vm}
+    Switch to vm    ${vm}
+    ${policy_cfg}    Current VM should have policy client enabled
+
+    ${rc}    Run Command    systemctl is-enabled givc-policy-init    return=rc    rc_match=skip
+    IF    ${rc}[0] != 0
```
fedorovpv, line 83: Not a mistake, but I think it could be nice to use 'Should not be equal' here

Robot-Framework/test-suites/security-tests/givc_policy.robot
```diff
@@ -0,0 +1,122 @@
...
+
+${POLICY_BACKUP_DIR}    /tmp
+
+
+*** Test Cases ***
+
+GIVC policy init on boot in policy-enabled VMs
+    [Documentation]  Verify givc-policy-init completed without known errors and created policy files.
+    [Tags]           SP-T365
+    Skip if VM list is empty    No reachable VMs with GIVC policy client enabled    @{POLICY_CLIENT_VMS}
```
milva-unikie: Why is the test skipped and not failed if GIVC policy client is not enabled?

Robot-Framework/test-suites/security-tests/givc_policy.robot
```diff
@@ -0,0 +1,122 @@
...
+    ${count}    Get Length    ${vm_list}
+    IF    ${count} == 0
+        SKIP    ${message}
+    END
+
+Current VM should have policy client enabled
+    ${policy_cfg}    Get current VM policy config
+    ${enabled}    Evaluate    bool($policy_cfg.get("enable"))
+    IF    not ${enabled}
+        SKIP    Policy client not enabled in current VM
```
milva-unikie: Same here, why skip and not fail?

Robot-Framework/test-suites/security-tests/givc_policy.robot
```diff
@@ -0,0 +1,122 @@
...
+Verify givc policy init in ${vm}
+    Switch to vm    ${vm}
+    ${policy_cfg}    Current VM should have policy client enabled
+
+    ${rc}    Run Command    systemctl is-enabled givc-policy-init    return=rc    rc_match=skip
+    IF    ${rc}[0] != 0
+        FAIL    givc-policy-init is not enabled in ${vm}
+    END
+
+    Service should not be failed    givc-policy-init
```
milva-unikie, line 87: The service not failing does not mean that it is active. It could for example be `inactive` or stuck at `activating`. Is it intentional that the test checks only for clean failures?

Robot-Framework/resources/service_keywords.resource
```diff
@@ -7,6 +7,13 @@ Resource            ../resources/ssh_keywords.resource
 
 
 *** Keywords ***
+Service should not be failed
+    [Arguments]    ${service}
+    ${result}    Run Command    systemctl show -p Result --value ${service}    rc_match=skip
+    ${state}    Run Command    systemctl show -p ActiveState --value ${service}    rc_match=skip
+    Should Not Contain    ${result}    failed
+    Should Not Contain    ${state}    failed
```
milva-unikie, line 20: Would be nice to save `journalctl` logs for debugging in case this keyword fails

Robot-Framework/test-suites/security-tests/givc_policy.robot
```diff
@@ -0,0 +1,122 @@
...
+        SKIP    Policy client not enabled in current VM
+    END
+    RETURN    ${policy_cfg}
+
+Get configured policy destination
+    [Arguments]    ${policy_cfg}    ${policy_name}
+    ${policies}    Evaluate    $policy_cfg.get("policies", {})
+    ${has_policy}    Evaluate    $policy_name in $policies
+    IF    not ${has_policy}
+        SKIP    Policy ${policy_name} not configured in current VM
```
milva-unikie: Same here, why skip and not fail?

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: givc-policy-init has "oneshot" type. So it should be enough to check that
  - it is enabled
  - it did not fail
  - its journal does not show the expected error pattern
  - expected effects exist on disk (which is the main focus of the test)

After boot it is expected to be inactive (dead)

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: Probably we should require that @{POLICY_CLIENT_VMS} includes at least business-vm and chrome-vm (current state). I will change it like that.

But the main idea behind that structure was to test according to maturity of ghaf, extend testing to only those vms which have givc policy.

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: Yes, at this point skipping is wrong. It should definitely fail. Changing it.

In setup it is already checked that givc policy is enabled for the vms which are selected for testing, so this check is kind of unnecessary. The keyword is mainly for getting policy config.

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: Thanks. This should also fail. Changing.

(However, this is a bit over-guarding because the policy_name argument actually comes to the keyword from policy_cfg.)

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: Ok. I think I will leave it. It looks cleaner like that for me.

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: Ended up removing this keyword and using just `Get current VM policy config` instead.

leivos-unikie:
Tested on Lenovo-X1 after the latest fixes:
https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6754/

Robot-Framework/test-suites/security-tests/givc_policy.robot
```diff
@@ -0,0 +1,116 @@
...
+        END
+        ${policy_cfg}    Evaluate    json.loads($config_raw).get("capabilities", {}).get("policy", {})    json
+        ${enabled}    Evaluate    bool($policy_cfg.get("enable"))
+        IF    not ${enabled}
+            CONTINUE
+        END
+        Append To List    ${POLICY_CLIENT_VMS}    ${vm}
+    END
+    List Should Contain Value    ${POLICY_CLIENT_VMS}    business-vm
+    List Should Contain Value    ${POLICY_CLIENT_VMS}    chrome-vm
```
milva-unikie: Is there a reason why to not use `${BUSINESS_VM}` and `${CHROME_VM}`?

Robot-Framework/test-suites/security-tests/givc_policy.robot
milva-unikie: This skips seems to now be redundant, `GIVC policy suite setup` will fail if VM list is empty

leivos-unikie: Yes. Removed it.

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: No. Changed to those variables.

Robot-Framework/test-suites/security-tests/givc_policy.robot
```diff
@@ -0,0 +1,115 @@
...
+    Set Suite Variable      @{VM_LIST_WITH_HOST}
+    Set Suite Variable      @{POLICY_CLIENT_VMS}
+
+Get current VM policy config
+    ${config_raw}    Run Command    cat /etc/givc-agent/config.json
+    ${config}    Evaluate    json.loads($config_raw)    json
+    ${policy_cfg}    Evaluate    $config["capabilities"]["policy"]
+    RETURN    ${policy_cfg}
+
+Skip if VM list is empty
```
milva-unikie: Sorry one last comment, this keyword is now not used

milva-unikie (approved):

Robot-Framework/test-suites/security-tests/givc_policy.robot
leivos-unikie: Thanks. Removed it.

PR #764, opened 2026-04-14
Recover after logging failure

milva-unikie:
- Execute update tests in a logged in state. This makes connecting to app VMs more reliable and aligns with other tests.
- Add a suite teardown for update tests. Updating is currently producing too much logs for the admin-vm to handle and causes the logging to Grafana stop.
  - `Update Teardown` stops the logging service, deletes extra logs and then restarts the service.
  - `Elevate to superuser` needs an extended timeout because admin-vm is working very slowly due to running out of space. 
- Confirm that logging is working before running `Check file system changes are logged` and `Check that system logs privilege uses and key events`. This will help with debugging.

Testrun https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/5432/

### DIFF
--- a/Robot-Framework/resources/ssh_keywords.resource
+++ b/Robot-Framework/resources/ssh_keywords.resource
@@ -309,13 +309,14 @@ Transfer Shell Script To VM
     Transfer Shell Script To DUT    performance-tests   ${script_name}   /tmp
 
 Elevate to superuser
+    [Arguments]    ${sleep}=0.5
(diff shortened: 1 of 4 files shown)


leivos-unikie (approved):
Good work!

As update can make such jam for centralized logging maybe we can suggest ghaf switching to some other less verbose logging mode for the time of update. Just a thought for the future, if that happens also with A/B update procedure.

PR #787, opened 2026-04-29
Add new test for access list

fedorovpv:
New test opens trusted browser, goes to the URL and looks for the text on the page.

yle.fi shouldn't be available, so test looks for the text of error.
HTTP and HTTPS have different errors by design.

Access list consist of 3 files, each file has multiple sections. Each URL for positive testing represents one section.

[Lenovo](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/5729/artifact/Robot-Framework/test-suites/SSRCSP-T362/log.html)
[Darter](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/5728/artifact/Robot-Framework/test-suites/SSRCSP-T362/log.html)

### DIFF
--- a/Robot-Framework/lib/GuiTesting.py
+++ b/Robot-Framework/lib/GuiTesting.py
@@ -18,12 +18,44 @@ def locate_image(screenshot, image, confidence):
     logging.info(image_center_in_mouse_coordinates)
     return image_center_in_mouse_coordinates
 
-def locate_text(screenshot, text, scale=1):
-    logging.info("Searching " + text)
-    image = Image.open(screenshot)
+def get_data_from_image(image, scale=1):
+    image = Image.open(image)
     image = image.resize((image.width * scale, image.height * scale), Image.BICUBIC)
     data = pytesseract.image_to_data(image, output_type=pytesseract.Output.DICT)
     logging.info(data)
+    return data
+
+def get_text(data: pytesseract.Output.DICT):
+    text_list = []
+    sentence = []
+
+    for word in data['text']:
+        if word == "":
+            if sentence:
+                text_list.append(' '.join(sentence))
+                sentence.clear()
+        else:
+            sentence.append(word)
+
+    if sentence:
+        text_list.append(' '.join(sentence))
+
+    logging.info(text_list)
+    return text_list
+
+def is_text_on_the_screen(screenshot, text, scale=1):
+    logging.info("Searching " + text)
+    data = get_data_from_image(screenshot, scale)
+
+    for sentence in get_text(data):
+        if text in sentence:
+            return True
+
+    return False
+
+def locate_text(screenshot, text, scale=1):
+    logging.info("Searching " + text)
+    data = get_data_from_image(screenshot, scale)
 
     # Loop through results to find matching text
     for i, word in enumerate(data['text']):
--- a/Robot-Framework/resources/app_keywords.resource
+++ b/Robot-Framework/resources/app_keywords.resource
@@ -20,7 +20,7 @@ ${APP_OUTPUT_FILE}   /home/${USER_LOGIN}/output.log
 Start application in VM
     [Documentation]    Use always_check_vm=True if calling from tests other than functional
     ...                After calling for this keyword use "Kill App in VM" in the test teardown
-    [Arguments]    ${app_name}   ${app-vm}   ${process_name}   ${always_check_vm}=False
+    [Arguments]    ${app_name}   ${app-vm}   ${process_name}   ${always_check_vm}=False    ${params_string}=${EMPTY}
     Set Test Variable    ${TEST_APP-VM}         ${app-vm}
     Set Test Variable    ${TEST_PROCESS_NAME}   ${process_name}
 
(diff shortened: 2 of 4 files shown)


Robot-Framework/test-suites/security-tests/network.robot
```diff
@@ -29,6 +31,27 @@ Check OpenSSL3 is Available In Nix Store
     ${output}        Run Command    ls /nix/store | grep openssl-3    rc_match=skip
     Should Not Be Empty    ${output}    msg=Found no openssl in Nix Store
 
+Check Access List In Trusted Browser
+    [Tags]    SSRCSP-T362   lenovo-x1  darter-pro
```
milva-unikie: Please use the short version of the tag
```suggestion
    [Tags]    SP-T362   lenovo-x1  darter-pro
```

Robot-Framework/test-suites/security-tests/network.robot
```diff
@@ -55,4 +78,19 @@ Check ip is in the blacklist
...
\ No newline at end of file
+    Run Command  ipset del f2b-sshBlacklist ${ip}   sudo=True
+
+Check Access List In Trusted Browser Template
+    [Documentation]    Running Trusted Browser and trying to find text on the opened page.
+    ...                If page is blocked by access list, error depends on the protocol used:
+    ...                   https: 'This site can’t be reached'
+    ...                   http:  'Access Denied'
+    [Arguments]    ${url}   ${text_to_find}
+    Start application in VM   "Trusted Browser"   ${BUSINESS_VM}   google-chrome    params_string=-- ${url}
```
milva-unikie:
```suggestion
    Start application in VM   "Trusted Browser"   ${BUSINESS_VM}   google-chrome    always_check_vm=True  params_string=-- ${url}
```
This should not cause issues with this case, but `Start application in VM` asks to call with `always_check_vm=True` when outside of functional tests. `always_check_vm=False` is there to optimize the functional tests that are constantly running.

milva-unikie (commented):
Some general comments:
1. I do think that this is a GUI test and should be moved under the GUI tests. As the person who usually rewrites them after changes, it is a lot easier to manage when they are in the same suite.
2. Please add the screen recording to this test case to help with debugging (`Start screen recording` & `Stop screen recording   ${TEST_STATUS}   ${TEST_NAME}`)
3. https://word.cloud.microsoft/, https://excel.cloud.microsoft/ and https://powerpoint.cloud.microsoft/ should be included to the pages and the related Zephyr tags added to the test case. To be able to remove the manual cases from the regression cycle we need to be able to link the corresponding automated case.

Robot-Framework/test-suites/security-tests/network.robot
fedorovpv: Does the short one lead to anywhere? 

I mean, I understand how to use the full one to get the link to the case, but how should I use short one without SSRC part?

milva-unikie: It does not lead to anywhere, but we use the short one for every single test case. In our previous Jira the tag was SP-T362 and after migration to the current one we decided not to change them to the format of SSRCSP-T362.

If we want to change them it should be done all at the same time.

fedorovpv:
>     1. I do think that this is a GUI test and should be moved under the GUI tests. As the person who usually rewrites them after changes, it is a lot easier to manage when they are in the same suite.

It's not a big issue for me, I can move it. But should we group tests by the way we test instead of _what we test_? As for me we are not testing GUI here, but the security

>     3. https://word.cloud.microsoft/, https://excel.cloud.microsoft/ and https://powerpoint.cloud.microsoft/ should be included to the pages and the related Zephyr tags added to the test case. To be able to remove the manual cases from the regression cycle we need to be able to link the corresponding automated case.

OK, I can add it. 

Now it uses templates, so tags for excel, powerpoint and words would be in one line.

```
`Check Access List In Trusted Browser
    [Tags]    SP-T362  SP-T_WORD SP-T_POWERPOINT  lenovo-x1  darter-pro`
    
```
    
But if there are separate cases I can copy-paste and make something like this

```
Check Word
    [Tags]   SP-T_WORD   lenovo-x1  darter-pro
    [Template]    Check Access List In Trusted Browser Template
    https://word.cloud.microsoft/    text_to_find=Welcome to Word

Check Powerpoint
    [Tags]    SP-T_POWERPOINT   lenovo-x1  darter-pro
    [Template]    Check Access List In Trusted Browser Template
     https://poerpoint.cloud.microsoft/    text_to_find=Welcome to Powerpoint

Check Access List In Trusted Browser
    [Tags]    SP-T362   lenovo-x1  darter-pro
    [Template]    Check Access List In Trusted Browser Template
    # Pages outside access list shouldn't be available via Trusted Browser. Http and https has different errors
    https://yle.fi                          text_to_find=This site can’t be reached
    ....

```

milva-unikie:
> It's not a big issue for me, I can move it. But should we group tests by the way we test instead of _what we test_? As for me we are not testing GUI here, but the security

I agree, but that is not how the tests are currently organized. Functional tests also have a lot of tests that could be categorized as security, but they are there so that it is easier to include them to the pre-merge & bat cycles. It would probably be good to discuss if we want to rearrange the tests, but that should be done all at the same time.

> Now it uses templates, so tags for excel, powerpoint and words would be in one line.

I think this is totally fine, we also have other tests that have multiple tags.

fedorovpv:
New run after fixes

[Lenovo](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/5781/artifact/Robot-Framework/test-suites/SP-T362/log.html)

milva-unikie (approved):

PR #808, opened 2026-05-13
Update skips and tags

milva-unikie:
- Remove app is running check from `Stop screen recording`, no need to check it before running `Kill process by name`. In the nightly there was an failure where screen recorder was running when `Stop screen recording` checked it but not when `Kill process by name` checked the same thing.
- Arrange tags with `organize_robot_tags.py`. Add new tags to the tag priority list.
- Update `Validate Forward Secure Sealing` skip to point to the new bug.
- Add a note about Falcon being removed to `list_of_skipped_tests.md`.

Lenovo X1 gui https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6038/

### DIFF
--- a/Robot-Framework/resources/gui_keywords.resource
+++ b/Robot-Framework/resources/gui_keywords.resource
@@ -354,13 +354,8 @@ Start screen recording
 
 Stop screen recording
     [Arguments]      ${test_status}   ${test_name}=""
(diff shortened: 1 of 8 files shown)


milva-unikie:
Rebased and updated the skip number for `journal-fss-verify.service` skip too

fedorovpv (approved):

azbeleva (approved):

PR #809, opened 2026-05-18
Add more power tests

milva-unikie:
* Add tests for `Log out and log in with shortcut` and `Suspend and wake up from lock screen`.
* Rename gui-power-menu tests to just gui-power tests.
* Rename old gui power tests tests with more descriptive name. Remove GUI from the start of test names, it is not used in any other tests.

Testruns
- [Darter Pro](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6066/)
- [Lenovo X1](https://ci-prod.vedenemo.dev/job/ghaf-hw-test-manual/1989/) (in prod with power measurement)

### DIFF
--- a/Robot-Framework/test-suites/gui-tests/gui_power.robot
+++ b/Robot-Framework/test-suites/gui-tests/gui_power.robot
@@ -2,8 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 *** Settings ***
-Documentation       Testing taskbar power widget options
-Test Tags           gui-power-menu
+Documentation       Testing power options
+Test Tags           gui-power
 
 Resource            ../../resources/common_keywords.resource
 Resource            ../../resources/device_control.resource
@@ -20,11 +20,8 @@ Test Teardown       GUI Power Test Teardown
 
 *** Test Cases ***
 
-GUI Suspend and wake up
-    [Documentation]   Suspend the device via GUI taskbar suspend icon.
-    ...               Check that the device is suspended.
-    ...               Wake up by pressing the power button for 1 sec.
-    ...               Check that the device is awake.
+Suspend and wake up from power menu
+    [Documentation]   Suspend the device via GUI taskbar suspend icon and wake up.
     ...               Logs device power consumption during the test
     ...               if power measurement tooling is set.
     [Tags]            SP-T75  SP-T75-3  lenovo-x1  darter-pro  lab-only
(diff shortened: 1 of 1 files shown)


Robot-Framework/test-suites/gui-tests/gui_power.robot
```diff
@@ -124,6 +101,25 @@ GUI Log out and log in
...
     Log in, unlock and verify
 
+Log out and log in with shortcut
+    [Documentation]   Logout via logout shortcut and verify logged out state.
+    ...               Login and verify that desktop is available.
+    [Tags]            SP-T186  lenovo-x1  darter-pro
+    Press Key(s)                LEFTMETA+LEFTSHIFT+ESC
+    Tab and enter               tabs=1
+    ${logout_status}            Check if logged out
+    IF  not ${logout_status}    FAIL  Logout failed.
```
fedorovpv: Why do we need if clause instead of [Should Be True](https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Should%20Be%20True)?

Robot-Framework/test-suites/gui-tests/gui_power.robot
```diff
@@ -124,6 +101,25 @@ GUI Log out and log in
...
+    Log in, unlock and verify
+
+Suspend and wake up from lock screen
+    [Documentation]   Suspend the device from lock screen suspend icon and wake up.
+    [Tags]            SP-T245  lenovo-x1  darter-pro
+    Press Key(s)          LEFTMETA+ESC
+    Locate on screen      image  ${LOCK_ICON}   0.95  10
```
fedorovpv, line 118: Just an idea, not asking to change the way we work now. 

But don't we want to have some screen(or window/panel) object were we could store coordinates of buttons and ways to interact with it? 

It could be more readable and reusable to have just some `Main Panel. Click Suspend Button`.

milva-unikie: This would be a good idea, but agreed that it is not in the scope of this PR.

For the power menu there is a keyword `Select power menu option`. The lock screen only has one button with actual functionality (the suspend button), so writing a keyword for it felt unnecessarily at this point.

(It also would be nice to not use hard coded coordinates, but I could not find the correct suspend icon. I have previously had this same issue with power menu icons too)

Robot-Framework/test-suites/gui-tests/gui_power.robot
milva-unikie: Only because I copied this part directly from `Log out and log in` and that test uses `IF` :) 

I can change them both, `Should Be True` would be cleaner.

fedorovpv (approved):

milva-unikie:
Testrun after changes https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6068/

PR #818, opened 2026-05-22
Add zephyr

fedorovpv:
Integration with the Jira Zephyr. 

Now it saves all the results for all the cases to the [same place](https://jira.tii.ae/secure/Tests.jspa#/testCase/SSRCSP-T114), because some cases have multiple tags and we need to decide ho we will use it.

After merging it will not be used immediately, we need to make PR to the infra with getting credentials and adding listener to the command that runs tests

This [Jenkins Run ](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6178/) has manually edited groovy script with getting creds and using listener. 

Where and how creds should be stored is discussed right now, but doesn't have a lot of influence to this code

### DIFF
--- a/Robot-Framework/lib/Zephyr.py
+++ b/Robot-Framework/lib/Zephyr.py
@@ -1,0 +1,125 @@
+# SPDX-FileCopyrightText: 2022-2026 Technology Innovation Institute (TII)
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import requests
+from robot.api import logger
+from http import HTTPStatus
+
+class ZephyrError(Exception):
+    pass
+
+class ZephyrTestStatuses:
+    """Container for supported Zephyr Scale Plugin test statuses"""
+    NOT_EXECUTED = 'Not Executed'
+    IN_PROGRESS = 'In Progress'
+    PASS = 'Pass'
+    FAIL = 'Fail'
+    BLOCKED = 'Blocked'
+
+class Zephyr:
+    """Class for integration with Zephyr Scale Plugin"""
+    connected: bool
+    zephyr_api: str
+    project_key: str
+    test_statuses = ZephyrTestStatuses
+
+    def __init__(self, project_key="SSRCSP"):
+        """ Initialize Zephyr Scale API client."""
+        self.token = os.getenv("JIRA_TOKEN")
+        self.jira_link = os.getenv("JIRA_LINK")
+
+        self.zephyr_api = f"{self.jira_link}/rest/atm/1.0"
+        self.project_key = project_key
+        headers = {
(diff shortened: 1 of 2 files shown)


leivos-unikie:
If test automation executes a test and logs the result in Zephyr I want to have that also logged who executed the test. Could it add "test-automation" or the name of the test agent at the "Executed by" label of the Zephyr test case?

It would be good to see also how this behaves in error cases, or do you already have some simulations on that? e.g. failing to connect to Zephyr.

Does it log anything if there are no errors? Maybe some log in init at check_connection would be nice, so it would not be so invisible. And if there are errors does it log too much?

fedorovpv:
@leivos-unikie 

> If test automation executes a test and logs the result in Zephyr I want to have that also logged who executed the test. Could it add "test-automation" or the name of the test agent at the "Executed by" label of the Zephyr test case?

In my opinion there is no limitations to do it. There is some env var in jenkins run with name of the person who started it, so we can forward it to the Zephyr. 

However, I don't I think that we need it at the moment. If there will be info about runs from our local machines during test development or some debug from ghaf-hw-test-manual it would add noise and it would be hard to understand where the actual error is. 

The main idea now is to save results from the stable version on the tests, _from the nightly runs_. If you have history of the stable runs you can analyze previous test failures, compare how tests behaved on different ghaf versions, etc. So, if it's about nightly runs only, it run automatically and if someone triggers it manually it just rerun of failed automated one.

But as I said in the first line: there is no technical limitation for that, I can add it if it's a deal-breaker for you. I just don't see any usecase outside nightly runs now, so I don't think it has high priority and it could be added later. 

----------------------
> It would be good to see also how this behaves in error cases, or do you already have some simulations on that? e.g. failing to connect to Zephyr.

**1. Wrong/expired token, wrong jira link**
image
image

**2. No jira link:**
image
image

**3. No token**
image
image

**4. No tag found**
image

**5. Non-existing test cycle** 
image

6. **Non-existing test case**
image

----------------------


> Does it log anything if there are no errors? Maybe some log in init at check_connection would be nice, so it would not be so invisible. 

Added line to the console if connections was OK
image

leivos-unikie:
> But as I said in the first line: there is no technical limitation for that, I can add it if it's a deal-breaker for you. I just
> don't see any usecase outside nightly runs now, so I don't think it has high priority and it could be added later.

At this point this is not critical, especially when only those specific cycles are used in integration. If in the future we can approach some kind of hybrid execution in REEM cycles (partially automated, partially manual) then it will be quite important.

Error logging looks ok. And good to have that "Succesfully connected..."

leivos-unikie (approved):

PR #825, opened 2026-05-28
GUI improvements

milva-unikie:
- Add iterations and a debug screenshot for `Check if locked` after suspension.
- Rename `Type string and press enter` to `Type string` with argument `enter_at_end`. Enter is only used half of the time, so I think this is cleaner format. Happy to hear other opinions though.
- Reduce the console logging in `Locate on screen` so that there is not a huge wall of text, especially after login.

**Testruns**
- Lenovo X1 [guiANDsuspension](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6263/)
- Darter Pro storeDisk [gui](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6264/)

### DIFF
--- a/Robot-Framework/resources/gui_keywords.resource
+++ b/Robot-Framework/resources/gui_keywords.resource
@@ -86,17 +86,13 @@ Log out with loginctl
     Run Command    loginctl terminate-user ${USER_LOGIN}   sudo=True
     [Teardown]   Switch to vm   ${GUI_VM}  user=${USER_LOGIN}
 
-Type string and press enter
-    [Arguments]    ${string}=${EMPTY}  ${enter}=True   ${sudo}=False
+Type string
+    [Arguments]    ${string}=${EMPTY}  ${enter_at_end}=False   ${sudo}=False
     Log To Console    Typing "${string}"
     IF  $string != '${EMPTY}'
         Run ydotool command   type ${string}  ${sudo}
     END
-    IF  ${enter}
-        Press Key(s)      ENTER   ${sudo}
-    ELSE
-        Log To Console    Skipping Enter
-    END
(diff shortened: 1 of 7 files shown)


Robot-Framework/resources/gui_keywords.resource
```diff
@@ -136,11 +132,11 @@ Locate on screen
     ${coordinates}=        Set Variable  ${EMPTY}
     ${locate_status}=      Set Variable  FAIL
     Run Command            mkdir -p test-images
+    Log To Console         Taking a screenshot to locate ${type} ${searched_item}   console=True   no_newline=True
     FOR   ${i}   IN RANGE  ${iterations}
-        Log To Console           Taking a screenshot (#${i+1})...  no_newline=True
         ${screenshot_status}  ${result}     Take a screenshot
+        Log To Console        .   no_newline=True
```
azbeleva: not sure, but wouldn't it be better to make counting like 1 2 3 4... instead of "..."
now in logs it is not obvious what is the difference between lines:
```
Taking a screenshot to locate image ghaf-close.png.......
Taking a screenshot to locate image lock.png...
Taking a screenshot to locate text LogOut.
```

azbeleva: but I am not sure how useful this information is, and with dots it is also fine for me

azbeleva (approved):

Robot-Framework/resources/gui_keywords.resource
milva-unikie: Oh yeah I can change that, counting would be more useful than just dots. I did not really think it more than to have something small showing that the screenshots are being taken.

milva-unikie:
New testrun after changes https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6268/

PR #859, opened 2026-06-17
Refactor Check If Device Up

fedorovpv:
### Goal
Reduce complexity of the `Check If Device Is Up` keyword.

### Main logic
I was trying to not reinvent the wheel, so I saved the way we check if device is up:
1. Check if device is pingable and SSH is UP
2. If variable ${UART_CAPTURE_ACTIVE} is False Check if Serial is UP, otherwise skip
3. Results of the checks are written to the global variables ${IS_AVAILABLE} and ${CONNECTION_TYPE} 

### Main things that were done:
1. Reduced the quantity of nested keywords, where it's possible, e.q. IF-THEN+FAIL -> Should Be
2. Replaced loops+time measuring with Wait Until Keyword Succeeds
3. Chunked code into keywords with readable names and logic like: 
      A. `KW that gives state` -> 
      B. `KW that checks state given by A` ->
      C. `KW that wait for the check B is passed`
4. Unified arguments  of the `Check If Device Is Up` keyword ${range} and ${timeout} to one ${retry}. It's the same as the WUKS has, and basically just passes argument to WUKS
5. Separated setting global variables from the main code of keyword. If laptop is available or not is not a constant thing and could change during the test run, that's why `${IS_AVAILABLE}` is always used with running `Check If Device Is Up` keyword, because you need to verify current value and install it before using it. Basically it works like returning value from the keyword, but with opportunity to forget to install proper value with `Check If Device Is Up`. It's one of the reason why using global variables outside case of storing constants is a [well-know anti-pattern
](https://stackoverflow.com/questions/18635136/why-are-global-variables-considered-bad-practice-node-js). I think we should reduce using it, it's not a huge problem yet, but of course it's not a goal of this task. Here I just separated it to make its removal easier in the future.

### Runs
|  | SSH + | SSH - |
| :----------- |:--------------:| :-------------:|
| **Serial +** | [ + +, all regression on darter](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6557/artifact/Robot-Framework/test-suites/darter-proANDregression/log.html)             | - +: [${UART_CAPTURE_ACTIVE} is False](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6578/artifact/Robot-Framework/test-suites/SP-T351-1/log.html#s1-s1-k1-k1-k1-k1-k2)                                           , [ ${UART_CAPTURE_ACTIVE} is True](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6569/artifact/Robot-Framework/test-suites/relaybootANDdarter-pro/log.html#s1-s1-s1-t1-k5-k3-k2-k1-k2-k1-k3) |
| **Serial -**      | [+ - ](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6579/artifact/Robot-Framework/test-suites/SP-T351-1/log.html#s1-s1-k1-k1-k1-k1-k1)           | [- -](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6571/artifact/Robot-Framework/test-suites/relaybootANDdarter-pro/log.html#s1-s1-s1-t1-k5-k3-k2-k1-k2-k2)   |

### Found issues
Check Serial never runs on the boot tests because of the ${UART_CAPTURE_ACTIVE} flag set to True
But check serial still could work if device passes boot test and after that reboots in the main section

### DIFF
--- a/Robot-Framework/resources/serial_keywords.resource
+++ b/Robot-Framework/resources/serial_keywords.resource
@@ -64,10 +64,10 @@ Check Serial Connection
 
     ${serial_status}=   Open Serial Port
     Should Not Be Equal    ${serial_status}    False    msg=Serial port is not available
-    Wait Until Keyword Succeeds    ${range}    1s    Check For Text Is In Serial Port    text=ghaf
+    Wait Until Keyword Succeeds    ${range}    1s    Check For Text In Serial Port    text=ghaf
     Delete All Ports
 
-Check For Text Is In Serial Port
+Check For Text In Serial Port
     [Arguments]    ${text}
     Write Data     ${\n}
     ${output}      SerialLibrary.Read Until
--- a/Robot-Framework/resources/serial_keywords.resource
+++ b/Robot-Framework/resources/serial_keywords.resource
@@ -57,29 +57,33 @@ Turn Relay On
 
 Check Serial Connection
     [Documentation]    Check if device is available by serial
-    [Arguments]        ${range}=30
+    [Arguments]        ${range}=30x
     Log To Console     Trying to connect via serial...
-    ${serial_status}   Open Serial Port
-    IF   ${serial_status}==False
-        Log       Serial port is not available      console=True
-        RETURN    False
-    END
-    FOR    ${i}    IN RANGE    ${range}
-        Write Data     ${\n}
-        ${output}      SerialLibrary.Read Until
-        ${status}      Run Keyword And Return Status    Should contain    ${output}    ghaf
-        IF    ${status}    BREAK
-        Sleep   1
-    END
+    Should Not Be True     ${UART_CAPTURE_ACTIVE}       msg=UART capture is active, can't check serial, data on the port could be lost.
+    Should Not Be Equal    ${SERIAL_PORT}      NONE     msg=There is no address for serial connection
+
+    ${serial_status}=   Open Serial Port
+    Should Not Be Equal    ${serial_status}    False    msg=Serial port is not available
+    Wait Until Keyword Succeeds    ${range}    1s    Check For Text Is In Serial Port    text=ghaf
     Delete All Ports
-    Set Global Variable    ${UART_CAPTURE_ACTIVE}    ${False}
-    IF  ${status}
-        Log To Console       Device is available via serial
-        Set Global Variable  ${CONNECTION_TYPE}    serial
-        Set Global Variable  ${IS_AVAILABLE}       True
-    ELSE
-        Log To Console       Device is not available via serial
+
+Check For Text Is In Serial Port
+    [Arguments]    ${text}
+    Write Data     ${\n}
+    ${output}      SerialLibrary.Read Until
+    Should contain    ${output}    ${text}
+
+Try To Check Serial On Device
+    [Arguments]    ${range}=30x
+    ${serial_status}=      Set variable    ${False}
+    TRY
+        Check Serial Connection     range=${range}
+        ${serial_status}=    Set Variable   ${True}
+        Log    Device is available via Serial Port.     console=True
+    EXCEPT   AS    ${exc}
+        Log    Device is not available via serial, failed with an exception:\n ${exc}!      console=True
     END
+    RETURN    ${serial_status}
 
 Log In To Ghaf OS
     [Documentation]       Log in with ${LOGIN} and ${PASSWORD}
--- a/Robot-Framework/resources/ssh_keywords.resource
+++ b/Robot-Framework/resources/ssh_keywords.resource
@@ -19,21 +19,25 @@ ${NETVM_SYNTAX}     ghaf-[0-9]{10}
 
 
 *** Keywords ***
-
 Ping Host
-    [Arguments]        ${hostname}  ${timeout}=${PING_SPACING}  ${allow_fail}=${False}
     [Documentation]    Ping the given hostname once and return boolean result
+    [Arguments]        ${hostname}  ${timeout}=${PING_SPACING}
     ${ping_output}=    Run Process   ping ${hostname} -c 1 -W ${timeout}   shell=True
-    ${ping_success}    Run Keyword And Return Status    Should Contain    ${ping_output.stdout}    1 received
-    IF  ${allow_fail} and not ${ping_success}
-        FAIL
-    END
-    Return From Keyword    ${ping_success}
+    ${received}=       Evaluate    '1 received' in $ping_output.stdout
+    RETURN    ${received}
 
-Check If Ping Fails
-    [Documentation]  Check that ping is not getting response from host
-    ${result}  Run Process  ping  ${DEVICE_IP_ADDRESS}  -c1  timeout=1s
-    Should Not Be Equal   ${result.rc}  ${0}
+Verify Ping Host
+    [Arguments]    ${hostname}    ${expected_status}=${True}   ${timeout}=${PING_SPACING}    ${console_log}=${False}
+    ${received_status}=    Ping Host            hostname=${hostname}    timeout=${timeout}
+    Run Keyword If         ${console_log}       Log      Ping ${hostname}... Status: ${received_status}     console=True
+    Should Be Equal        ${received_status}   ${expected_status}
+
+Wait For Device Going Offline
+    [Arguments]    ${hostname}   ${timeout}=15s     ${ping_timeout}=${PING_SPACING}
+    ${log_message}=     Set Variable      Device failed to go offline: still pingable after ${timeout}.
+    Wait Until Keyword Succeeds    ${timeout}  2s  Verify Ping Host     hostname=${hostname}    expected_status=${False}    timeout=${ping_timeout}
+    ${log_message}=     Set Variable      Device is not pingable: offline.
+    [Teardown]    Log     ${log_message}    console=True
 
 Get Ethernet Interface By Type
     [Documentation]    Returns ethernet interface filtered by usb or non-usb.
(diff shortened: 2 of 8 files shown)


Robot-Framework/resources/serial_keywords.resource
```diff
@@ -57,28 +57,33 @@ Turn Relay On
...
-        Log To Console       Device is not available via serial
+    Should Not Be True     ${UART_CAPTURE_ACTIVE}       msg=UART capture is active, can't check serial, data on the port could be lost.
+    Should Not Be Equal    ${SERIAL_PORT}      NONE     msg=There is no address for serial connection
+
+    ${serial_status}=   Open Serial Port
+    Should Not Be Equal    ${serial_status}    False    msg=Serial port is not available
+    Wait Until Keyword Succeeds    ${range}    1s    Check Text Is In Serial Port    text=ghaf
+    [Teardown]    Delete All Ports
+
+Check Text Is In Serial Port
```
milva-unikie: This keyword name is a bit confusing, it is reading serial output not the port. I would rename to `Check For Test In Serial Output` or something like that.

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -19,21 +19,24 @@ ${NETVM_SYNTAX}     ghaf-[0-9]{10}
 
 
 *** Keywords ***
-
 Ping Host
-    [Arguments]        ${hostname}  ${timeout}=${PING_SPACING}  ${allow_fail}=${False}
-    [Documentation]    Ping the given hostname once and return boolean result
```
milva-unikie, line 25: Why was the documentation removed?

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -263,30 +261,30 @@ Check if ssh is ready on vm
...
-        ${diff}=    Evaluate    int(time.time()) - int(${start_time})
-        IF   ${diff} > ${timeout}
-            BREAK
-        END
-    END
-    IF   ${is_ready} == False
-        Log To Console  Port 22 (ssh) of ${DEVICE} is not ready after ${timeout}
+Get Ssh Status
+    [Arguments]    ${ip_address}
+    ${rc}  	${output}	 Run And Return Rc And Output   nc -zvw3 ${ip_address} 22
```
milva-unikie, line 268: (Side note: [Run And Return Rc And Output](https://robotframework.org/robotframework/latest/libraries/OperatingSystem.html#Run%20And%20Return%20Rc%20And%20Output) is deprecated, should be changed at some point in the future)

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -263,30 +261,30 @@ Check if ssh is ready on vm
...
+    IF    ${rc} == 0
+        RETURN    ${True}   ${output}
+    ELSE
+        RETURN    ${False}  ${output}
     END
 
-    RETURN  ${is_ready}
+Check Ssh Status
+    [Arguments]    ${ip_address}    ${connection_succeeded}=${True}
+    ${status}    ${output}=         Get Ssh Status    ip_address=${ip_address}
```
milva-unikie: [Just my opinion] Is there a benefit for having `Get Ssh Status` as a separate keyword? For me having to jump to another keyword to check one line makes reading the code a lot more difficult. We are already quite a many keywords deep from `Check If Device Is Up`.

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -263,30 +261,30 @@ Check if ssh is ready on vm
...
+    Should Be Equal    ${status}         ${connection_succeeded}
+    RETURN    ${output}
+
+Wait Until Ssh Is Ready On Device
+    [Arguments]    ${ip_address}    ${timeout}=50s
+    [Timeout]    ${timeout}
+    ${is_ready}    Set Variable    ${False}
+    ${output}=    Wait Until Keyword Succeeds    ${timeout}   ${PING_SPACING}    Check Ssh Status    ip_address=${ip_address}
+    ${is_ready}    Set Variable    ${True}
```
milva-unikie, line 291: This won't work in a teardown, `${is_ready}` will always be set to `${True}` because [continue on failure](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#continue-on-failure) is enabled by default

Robot-Framework/test-suites/boot-test/relay_boot_test.robot
```diff
@@ -95,12 +95,8 @@ Turn OFF Device
...
         Turn Off Power
     END
-    ${device_not_available}  Run Keyword And Return Status  Wait Until Keyword Succeeds  30s  2s  Check If Ping Fails
-    IF  ${device_not_available} == True
-        Log To Console    Device is down
-    ELSE
-        Log To Console    Device is UP after the end of the test.
-    END
+
+    Wait For Device Going Offline    hostname=${DEVICE_IP_ADDRESS}    timeout=30s
```
milva-unikie, line 100: We have previously decided not to fail tests if the device does not shut down. I don't remember the exact reason but there was some device that sometimes did not shut down cleanly. Maybe it is time to try letting it fail, we will quickly see if it becomes an issue.

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -19,21 +19,24 @@ ${NETVM_SYNTAX}     ghaf-[0-9]{10}
...
-
-Check If Ping Fails
-    [Documentation]  Check that ping is not getting response from host
-    ${result}  Run Process  ping  ${DEVICE_IP_ADDRESS}  -c1  timeout=1s
-    Should Not Be Equal   ${result.rc}  ${0}
+    ${received}=       Evaluate    '1 received' in $ping_output.stdout
+    RETURN    ${received}
+
+Verify Ping Host
+    [Arguments]    ${hostname}    ${host_reached}=${True}   ${timeout}=${PING_SPACING}    ${console_log}=${False}
```
milva-unikie: Why is the argument called `${host_reached}`? `${expected_status}` or something similar would make a lot more sense to me.

Robot-Framework/resources/serial_keywords.resource
```diff
@@ -57,28 +57,33 @@ Turn Relay On
...
-        Log To Console       Device is available via serial
-        Set Global Variable  ${CONNECTION_TYPE}    serial
-        Set Global Variable  ${IS_AVAILABLE}       True
-    ELSE
-        Log To Console       Device is not available via serial
+    Should Not Be True     ${UART_CAPTURE_ACTIVE}       msg=UART capture is active, can't check serial, data on the port could be lost.
+    Should Not Be Equal    ${SERIAL_PORT}      NONE     msg=There is no address for serial connection
+
+    ${serial_status}=   Open Serial Port
+    Should Not Be Equal    ${serial_status}    False    msg=Serial port is not available
```
milva-unikie, line 66: This won't work in a teardown, it will always continue even if UART capture would be active or port not defined or port could not be opened. [Continue on failure](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#continue-on-failure) is enabled by default in teardowns.

Robot-Framework/resources/ssh_keywords.resource
```diff
@@ -116,48 +119,43 @@ Check Network Availability
...
+        Wait Until Ssh Is Ready On Device   ${DEVICE_IP_ADDRESS}
+
+        ${ssh_status}=      Set Variable    ${True}
+        Log To Console    Device is available via SSH
+    EXCEPT   AS    ${exc}
+        Log To Console    Device is not available via SSH, failed with an exception:\n ${exc}!
     END
+    RETURN    ${ssh_status}
+
+Set Up Global Variables If Device Is Up
```
milva-unikie: Why is this called `If Device Is Up`? Variables are set even when the device is not up

milva-unikie (changes requested):
I like this change, great job.

A few general comments in addition to the ones below:

1. Seconds and retries are now mixed when calling `Check If Device Is Up`. The execution time of one retry can vary significantly depending if the UART capture is on (it adds extra 5 seconds to every loop). I would change all long timeouts to use seconds. For `retry=5x` I think it makes sense to keep the iterations like they are.
2. `Log To Console` does not log the variables to the report, `Log <message> console=True` would do that. That is why we usually prefer `Log`.

Robot-Framework/test-suites/boot-test/relay_boot_test.robot
azbeleva: I was also thinking about that, and not totally sure, I would agree to see what will happen if we start failing them.

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: Yeah, it's more like "set up global variables FOR KEYWORD if device is up", but shortened. 

Agree, that naming is not the best here. Any suggestions?

Robot-Framework/resources/serial_keywords.resource
fedorovpv: Yeah, the idea was that we run Delete All Ports in teardown anyways, but looks like in original it was only after reading. So looks like it should be just after wuks without teardown

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: Because host could be reached like 1 or not like 0, and status is more like "passed", "reached" or something. But not a big deal for me, I can rename it

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: Ooops

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: Usually having separate keywords for:
1. Getting status of something
2. Comparing actual status with expected
3. Waiting for status becoming expected

helps to keep code readable and also allows reuse (1) if you don't need checks, just status. Also it's a really great practice for complicated checks(multiple statuses to check and/or multiple if clauses tied to results)

BUT here

**Readablity:**
`Get Ssh Status` is more clear than just `${rc}  	${output}	 Run And Return Rc And Output   nc -zvw3 ${ip_address} 22`. Of course this time check is simple and if kw `Check Ssh Status` has only one line that returns something for check, so it's pretty obvious what's going on here.

**Reuse:**
`Get Ssh Status` is used just once and it's in `Check Ssh Status`

So in fact, there is no big issue for me to move getting to check, I did it when reorganized keyword and was looking for abstractions, that could be wrapped in the keywords.

 If `get..` would be reused we can cut out it from the `check` anytime.

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: Here is it False, because 1.1.1.1 doesn't have open ssh port

Image

Robot-Framework/resources/ssh_keywords.resource
milva-unikie: `Set Global Device Availability Variables`?

Robot-Framework/resources/ssh_keywords.resource
milva-unikie: Oh right, I did not realize that `Wait Until Ssh Is Ready On Device` was called from `TRY-EXCEPT`. It should work fine then, thanks for digging up the logs.

Robot-Framework/resources/serial_keywords.resource
fedorovpv: fxd

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: fxd

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: yeap, it was just copied as is, so no fix for a while

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: Moved to one kw

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: renamed

Robot-Framework/resources/serial_keywords.resource
fedorovpv: removed teardown kw, delete all ports goes like a regular kw

Robot-Framework/resources/ssh_keywords.resource
fedorovpv: renamed

fedorovpv:
[log.html](https://github.com/user-attachments/files/29376844/log.html)
new run

Robot-Framework/resources/serial_keywords.resource
```diff
@@ -57,29 +57,33 @@ Turn Relay On
...
     Delete All Ports
-    Set Global Variable    ${UART_CAPTURE_ACTIVE}    ${False}
-    IF  ${status}
-        Log To Console       Device is available via serial
-        Set Global Variable  ${CONNECTION_TYPE}    serial
-        Set Global Variable  ${IS_AVAILABLE}       True
-    ELSE
-        Log To Console       Device is not available via serial
+
+Check For Text Is In Serial Port
```
milva-unikie: Extra `Is` in the keyword name?

milva-unikie (approved):

PR #870, opened 2026-06-30
Add a test for GUI lockout after failed login

milva-unikie:
Add a GUI version for `Account lockout after failed login`. Test tries to use wrong password 5 times in a row when logging in via GUI and then checks that the account is locked.

Screen recording does not work on the login screen so no video of this test. This could cause issues if the test is flaky, but in my local testing it has been very stable.

[Lenovo X1](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6749/artifact/Robot-Framework/test-suites/lenovo-x1ANDguiNOTstoreDisk-only/report.html#totals)
[Darter Pro](https://ci-dev.vedenemo.dev/job/ghaf-hw-test-manual/6750/artifact/Robot-Framework/test-suites/darter-proANDguiNOTstoreDisk-only/report.html#totals) (`Port already exists.` failed in the nightly too)

### DIFF
--- a/Robot-Framework/resources/gui_keywords.resource
+++ b/Robot-Framework/resources/gui_keywords.resource
@@ -33,13 +33,7 @@ Log in, unlock and verify
     ${logout_status}    Check if logged out   1
     IF  ${logout_status}
         Log in via GUI
-        ${status}    ${output}    Run Keyword And Ignore Error    Wait Until Keyword Succeeds    60s    1s    Verify login session
-        IF    '${status}' == 'FAIL'
-            Log To Console    Login session verification failed, retrying GUI login
-            Run Command       logger --priority=user.info "ROBOT_LOG: Login failed on ${DEVICE_TYPE}, job ${JOB}"
-            Log in via GUI
-            Wait Until Keyword Succeeds    60s    1s    Verify login session
-        END
+        Wait Until Keyword Succeeds    10s    1s    Verify login session
     ELSE
         ${lock}       Check if locked
         IF  ${lock}   Unlock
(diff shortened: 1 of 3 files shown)


leivos-unikie:
Teardown assumes that account was locked and performs another deliberate bad login during cleanup. I am just thinking is this good if the test actually failed to lock the account? Probably it will recover fine but can leave one faillock state behind.

Could the teardown procedure be different depending if “account was locked” or “account was never locked”?

That feature "first login after unlocking the account fails" is interesting, or is it a bug?

milva-unikie:
> Teardown assumes that account was locked and performs another deliberate bad login during cleanup. I am just thinking is this good if the test actually failed to lock the account? Probably it will recover fine but can leave one faillock state behind.
> Could the teardown procedure be different depending if “account was locked” or “account was never locked”?

Faillock bans are all wiped after successful login so I think it is fine to leave as is. The state is reset right before the one intentionally failing login, there are still plenty of attempts left.

> That feature "first login after unlocking the account fails" is interesting, or is it a bug?

Probably, but I don't think anyone will be interested in fixing it. The account is not supposed to get locked ideally.

leivos-unikie:
> > Teardown assumes that account was locked and performs another deliberate bad login during cleanup. I am just thinking is this good if the test actually failed to lock the account? Probably it will recover fine but can leave one faillock state behind.
> > Could the teardown procedure be different depending if “account was locked” or “account was never locked”?
> 
> Faillock bans are all wiped after successful login so I think it is fine to leave as is. The state is reset right before the one intentionally failing login, there are still plenty of attempts left.
> 
> > That feature "first login after unlocking the account fails" is interesting, or is it a bug?
> 
> Probably, but I don't think anyone will be interested in fixing it. The account is not supposed to get locked ideally.

Ok.

leivos-unikie (approved):

PR #924, opened 2026-08-09
fix(login): fix the test for app panel visible

brianmcgillion:


azbeleva:
What motivated this change? And what it was tested on?
I have also a concern, that text recognition is not very reliable, as a fallback it is okay, but primary method could be an image.

